PR #9, opened 2026-05-16
feat(sparse-moe): pipeline-parallel inference across N ranks

t8:
## Summary
- Splits the K2.6 SparseMoEEngine across N machines via tahoma-transport's existing TCP relay.
- New `dist.rs` defines a 3-frame wire protocol (Forward, Reset, Token) tailored to MoE pipeline parallelism — distinct from `ov-dist-spec`'s frame codes so a cross-engine mis-connect fails fast.
- `Runner` factored into `embed_layer0_tail` / `forward_shells` / `forward_head_last` so each rank only loads + runs the slice it owns. Single-stage path preserved via `LayerRange::full()`.
- `SparseMoEEngine::step` branches on rank: rank 0 drives prefill+decode; mid ranks forward and relay; last rank runs head + samples greedily and ships the token upstream.
- CLI drops the `--total 1` restriction on `sparse-moe`; even MoE-layer split is derived from `--rank/--total` when ShardSpec leaves the bounds at zero.

## Design decisions (locked with the user)
- **Control-channel back to rank 0** — last rank samples + sends token_id back over the existing upstream socket. Cheaper than re-broadcasting prompt+token (no quadratic prefill on subsequent tokens).
- **Modular `--total`** — engine supports any N >= 1, not hardcoded to 2.
- **Layer 0 KV punted** — stays stateless; long prefixes will grow quadratically on rank 0 until a stateful layer-0 IR replaces it. Follow-up PR.

## What's NOT in scope
- **Distributed sampling**: last rank uses argmax; per-frame temperature/top-p plumbing is a follow-up.
- **End-to-end 2-box eval on cascadia**: deferred to the next session — needs deploying the K2.6 artifacts to two AI PCs.
- **Topology-aware split**: even split for now; the design memo flagged latency-aware placement as a v2 question.

## Test plan
- [x] `cargo test -p tahoma-engine-sparse-moe` — 5 wire-protocol round-trips + 12 unit tests + 5 smoke tests, all green.
- [x] `cargo build --workspace` — clean.
- [x] `cargo fmt --all -- --check` — clean.
- [x] `cargo clippy -p tahoma-engine-sparse-moe --all-targets` — no new warnings (existing PR #7 lints unchanged).
- [ ] **2-box quality eval on cascadia-n00 + n01** — full 3-prompt eval from `quality_eval_20260516.jsonl` (Paris / Pacific / four). To run before merge.

## Stack note
This PR is stacked on #7 (`feat/sparse-moe-k26`) because the SparseMoEEngine itself is still in review there. When #7 lands, this branch will rebase onto main.

## Files
- `crates/tahoma-engine-sparse-moe/src/dist.rs` — new, 173 LOC (wire protocol)
- `crates/tahoma-engine-sparse-moe/src/engine.rs` — +550 net (rank-aware dispatch)
- `crates/tahoma-engine-sparse-moe/src/runner.rs` — +90 net (factored primitives)
- `crates/tahoma-engine-sparse-moe/tests/dist_wire.rs` — new, 158 LOC (integration tests)
- `crates/tahoma-cli/src/lib.rs` — −5 (drop total!=1 rejection, plumb with_rank)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2342,6 +2342,7 @@ dependencies = [
  "tahoma-engine",
  "tahoma-int4-gemm",
  "tahoma-ov-genai-shim",
+ "tahoma-transport",
  "tahoma-types",
  "tempfile",
  "thiserror 1.0.69",
--- a/crates/tahoma-cli/src/lib.rs
+++ b/crates/tahoma-cli/src/lib.rs
@@ -316,10 +316,8 @@ fn build_builder(args: &WorkerArgs) -> Result<Box<dyn Builder>> {
             }
         }
         EngineKind::SparseMoe => {
-            if args.total != 1 {
-                return Err(anyhow!("sparse-moe is single-stage only; use --total 1"));
-            }
-            let mut cfg = SparseMoEBuilderConfig::new(&args.model, &args.device);
+            let mut cfg = SparseMoEBuilderConfig::new(&args.model, &args.device)
+                .with_rank(args.rank, args.total);
             if let Some(dir) = &args.ov_cache_dir {
                 cfg.cache_dir = Some(dir.clone());
             }
--- a/crates/tahoma-engine-sparse-moe/Cargo.toml
+++ b/crates/tahoma-engine-sparse-moe/Cargo.toml
@@ -15,6 +15,7 @@ tahoma-types = { workspace = true }
 tahoma-engine = { workspace = true }
 tahoma-ov-genai-shim = { workspace = true }
 tahoma-int4-gemm = { workspace = true }
+tahoma-transport = { workspace = true }
 async-trait = { workspace = true }
 futures = { workspace = true }
 tokio = { workspace = true }
--- a/crates/tahoma-engine-sparse-moe/examples/dist_check.rs
+++ b/crates/tahoma-engine-sparse-moe/examples/dist_check.rs
@@ -1,0 +1,161 @@
+//! Cross-host wire-protocol smoke test.
+//!
+//! Spawns either the server side (binds, recv Forward, send Token) or
+//! the client side (connect, send Forward N times with synthetic
+//! hidden state, recv Token, time the round-trip). Used to verify the
+//! sparse-MoE pipeline-parallel wire protocol works over a real
+//! cross-host network (e.g. Tailscale's DERP relay between two Intel
+//! AI PCs).
+//!
+//! ```text
+//! # On the box that plays "last rank":
+//! cargo run --release --example dist_check -- server 9100
+//!
+//! # On the box that plays "rank 0":
+//! cargo run --release --example dist_check -- client <peer-ts-ip> 9100 100
+//! ```
+
+use std::env;
+use std::sync::Arc;
+use std::time::Instant;
+
+use tahoma_engine_sparse_moe::dist::{
+    recv_forward_body_server, recv_kind_client, recv_kind_server, recv_token_body_client,
+    send_forward, send_reset, send_token_upstream, FrameKind,
+};
+use tahoma_transport::{ActivationClient, ActivationServer};
+use tokio::sync::Mutex;
+
+const HIDDEN_SIZE: usize = 7168;
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let args: Vec<String> = env::args().collect();
+    let mode = args.get(1).map(String::as_str).unwrap_or("");
+    match mode {
+        "server" => {
+            let port: u16 = args.get(2).and_then(|s| s.parse().ok()).unwrap_or(9100);
+            run_server(port).await?;
+        }
+        "client" => {
+            let peer = args.get(2).cloned().ok_or("missing peer host")?;
+            let port: u16 = args.get(3).and_then(|s| s.parse().ok()).unwrap_or(9100);
+            let rounds: u32 = args.get(4).and_then(|s| s.parse().ok()).unwrap_or(10);
+            run_client(&peer, port, rounds).await?;
+        }
+        _ => {
+            eprintln!("usage:");
+            eprintln!("  dist_check server <bind-port>");
+            eprintln!("  dist_check client <peer-host> <peer-port> [rounds=10]");
+            std::process::exit(2);
+        }
+    }
+    Ok(())
+}
+
+async fn run_server(port: u16) -> Result<(), Box<dyn std::error::Error>> {
+    let mut server = ActivationServer::new("0.0.0.0", port);
+    server.start().await?;
+    println!("[server] listening on 0.0.0.0:{port}");
+    server.accept().await?;
+    println!("[server] peer accepted");
+    let server = Arc::new(Mutex::new(server));
+
+    let mut frame_count = 0u32;
+    let mut got_reset = false;
+    loop {
+        let kind = recv_kind_server(&server).await?;
+        let Some(kind) = kind else {
+            println!("[server] upstream closed cleanly after {frame_count} frames");
+            break;
+        };
+        match kind {
+            FrameKind::Reset => {
+                got_reset = true;
+                println!("[server] recv RESET");
+            }
+            FrameKind::Forward => {
+                let (past_seq_len, hidden, shape) = recv_forward_body_server(&server).await?;
+                let mean: f32 = hidden.iter().copied().sum::<f32>() / hidden.len() as f32;
+                frame_count += 1;
+                // Synthesise a "sampled" token from the input — XOR of
+                // past_seq_len with the hidden mean's bit pattern.
+                let token = (past_seq_len as i64) ^ (mean.to_bits() as i64);
+                send_token_upstream(&server, token).await?;
+                if frame_count <= 3 || frame_count.is_power_of_two() {
+                    println!(
+                        "[server] frame {frame_count} past_seq_len={past_seq_len} shape={:?} mean={:.4} → token={token}",
+                        shape, mean
+                    );
+                }
+            }
+            FrameKind::Token => {
+                println!("[server] unexpected TOKEN from upstream — ignoring");
+            }
+        }
+    }
+    println!(
+        "[server] frames processed: {frame_count} (got_reset={got_reset})"
+    );
+    Ok(())
+}
+
+async fn run_client(
+    peer: &str,
+    port: u16,
+    rounds: u32,
+) -> Result<(), Box<dyn std::error::Error>> {
+    println!("[client] connecting to {peer}:{port}");
+    let t_connect = Instant::now();
+    let mut client = ActivationClient::new(peer, port);
+    client
+        .connect_with_timeout(std::time::Duration::from_secs(30))
+        .await?;
+    let connect_ms = t_connect.elapsed().as_secs_f64() * 1000.0;
+    println!("[client] connected in {connect_ms:.1} ms");
+    let client = Arc::new(Mutex::new(client));
+
+    send_reset(&client).await?;
+    println!("[client] sent RESET");
+
+    let mut hidden = vec![0.0f32; HIDDEN_SIZE];
+    let shape = [1u32, 1, HIDDEN_SIZE as u32];
+    let mut rt_ms: Vec<f64> = Vec::with_capacity(rounds as usize);
+
+    let total_t0 = Instant::now();
+    for round in 0..rounds {
+        // Fill hidden with a deterministic pattern that varies per round.
+        for (i, v) in hidden.iter_mut().enumerate() {
+            *v = (i as f32) * 0.0001 + (round as f32) * 0.01;
+        }
+        let t0 = Instant::now();
+        send_forward(&client, round, &hidden, shape).await?;
+        let kind = recv_kind_client(&client).await?;
+        let Some(FrameKind::Token) = kind else {
+            return Err(format!("expected TOKEN, got {kind:?}").into());
+        };
+        let token = recv_token_body_client(&client).await?;
+        let dt = t0.elapsed().as_secs_f64() * 1000.0;
+        rt_ms.push(dt);
+        if round < 3 || round.is_power_of_two() {
+            println!(
+                "[client] round {round} rt={dt:.1} ms token={token}"
+            );
+        }
+    }
+    let total_ms = total_t0.elapsed().as_secs_f64() * 1000.0;
+    rt_ms.sort_by(|a, b| a.partial_cmp(b).unwrap());
+    let p50 = rt_ms[rt_ms.len() / 2];
+    let p99 = rt_ms[(rt_ms.len() * 99) / 100];
+    let avg = rt_ms.iter().sum::<f64>() / rt_ms.len() as f64;
+    let bytes_per_forward = 8 /* kind+past */ + 20 /* tensor header */ + HIDDEN_SIZE * 4 /* f32 */;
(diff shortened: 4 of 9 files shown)


t8:
## Cross-host wire-protocol validation on cascadia AI PCs

Ran the new `dist_check` example (commit ecf785e) across two Intel Lunar Lake AI PCs on the cascadia tailnet:

- **Server**: cascadia-matias-03 (100.123.40.123) — Intel Core Ultra 7 258V, 32 GB RAM, Win 11
- **Client**: cascadia-matias-02 (100.77.178.45) — same hardware
- **Path**: Tailscale DERP relay via `sea` (direct UDP between AI PCs is blocked per Intel network config; documented in cascadia-fleet's INTEL_TIBER.md)

Both boxes built `cargo build --release -p tahoma-engine-sparse-moe --example dist_check` in ~75 s with the gnu toolchain. No code changes needed for Windows.

### Results — 100 Reset+Forward+Token round trips with K2.6-sized hidden state (1×1×7168 f32, 28,712 B/round)

| metric | value |
|---|---|
| connect | 20.5 ms |
| per-round avg | **22.4 ms** |
| per-round p50 | 19.5 ms |
| per-round p99 | 102.0 ms (first round) |
| total (100 rounds) | 2236 ms |
| wire bytes/round | 28,712 (28,700 Forward + 12 Token) |

### What this proves

1. The wire protocol works correctly across a non-trivial network (DERP relay, ~16 ms baseline RTT, two NAT'd /24 networks).
2. The K2.6-sized 28 KB hidden state moves in ~22 ms steady-state — the transport is **not** the throughput bottleneck for any plausible compute time. At K2.6's expected ~0.1 tok/s compute, transport adds <1% overhead.
3. Per-frame past_seq_len, hidden mean, and synthesised token agree byte-for-byte between client and server logs — the F32 encoding, big-endian framing, and tensor header all match the dist.rs spec.

### What's NOT validated

- **Full K2.6 quality eval on 2 boxes**: needs the 553 GB model transferred to both cascadia boxes (~275 GB per box at the sparse-MoE split). At DERP relay throughput, the transfer is a multi-hour task by itself, and OpenVINO Genai's C++ dev headers aren't installed on the cascadia boxes either — separate ops effort.
- **Direct UDP path**: Intel's network blocks WireGuard UDP between AI PCs; all inter-node traffic goes through DERP relay. Same conditions any future inference deploy would face.

Code-complete for the design we agreed on; the 3-prompt K2.6 eval crossover stays as a follow-up.

t8:
## K2.6 output divergence — investigation results

Tested on miner with the fixed K2.6 tokenizer (rainier `fa5d41f`), same prompt "The capital of France is", max 4 tokens, temp 0, rep_penalty 1.05:

| Path | Output | Notes |
|---|---|---|
| Python prototype `--rust-shells --rust-shells-int4` | `' Paris. The E'` | Reference (correct) |
| Tahoma Rust engine (this PR, `sparse-moe` engine) | `' a very very very'` | Diverges |
| Tahoma engine **with warmup disabled** | `' a very very very'` | Same — rules out warmup state leak |
| Python prototype **without** `--rust-shells` (OV shells, Rust experts) | _in flight_ | Will pin OV-shells-vs-Rust-shells |

### What's ruled out
- Tokenization (byte-identical between paths on 5 verified prompts)
- Sampling (Python `sample_token` and Rust `sample` are line-by-line equivalents)
- Layer 0 / head IR (same path, same input)
- Mask construction (`causal_mask` identical)
- Input names/dtypes/shapes (verified against shell IR XML)
- Output index resolution (engine asserts all 7 canonical outputs)
- Expert dispatch (same `tahoma_int4_expert_forward` cdylib call)
- Warmup state leak (confirmed via no-warmup re-run)

### Pinned to:
The **shell forward pass**. Tahoma engine drives OV shell IRs via `tahoma-ov-genai-shim`. The Python reference path with `--rust-shells` BYPASSES OV for shells entirely and runs `tahoma_int4_shell_forward_int4` from the cdylib. Same weights, different evaluator → different outputs.

### Next step (after the in-flight OV-only Python test confirms)
- If Python-OV-shells ALSO diverges → OV shell IRs have a numerical bug exposed by either dynamic-shape or bf16-input handling. Fix path: instrument both Rust engine and Python OV path to dump `attn_out_post_norm` first-8 floats after shells 1/30/60, find first divergent layer, look at OV plugin behavior on that subgraph.
- If Python-OV-shells matches "Paris" → bug is specific to the tahoma engine's invocation of OV shells (not OV itself). Likely candidates: bf16 set_tensor handling in `shim.cpp:set_input`, or stale output buffer aliasing.

Full memory at `~/.claude/.../memory/k26_output_divergence.md`. Engine path works end-to-end; quality eval is gated on resolving this.

t8:
## Update: bug pinned — OV shell IRs are at fault, not the tahoma engine

Smoking-gun test (Python prototype WITHOUT \`--rust-shells\`, i.e. using OV shells but Rust experts — same configuration as the tahoma engine) completed:

\`\`\`
Output: ' a very very very'
\`\`\`

That's the **same** wrong output as the tahoma Rust engine, token-for-token (\`step 0: token=261=' a'\`, then \`token=2098=' very'\` for the next 3 decode steps).

### What this proves
The tahoma engine's invocation of OV shells is faithful — same input, same output as the Python OV path. The bug is in the **OV shell IRs themselves** (\`kimi-k26-shells-kv/layer_NN/openvino_model.{xml,bin}\`).

Rainier's eval used \`--rust-shells\` which BYPASSES the OV shell evaluator entirely (parses the IR for weight tensors, then runs \`tahoma_int4_shell_forward_int4\` from the cdylib). That path is known-correct (produces "Paris"). So the OV shell IRs were never validated end-to-end; the bug has been latent since they were exported.

### Fix options
1. **Re-export shells correctly** in rainier — `scripts/export_cached_shards_v5.py` (or whichever script produced the shells-kv tree) has a bug, probably RoPE-related given the K2.6 model-quirks history. The correct math is in `tahoma-int4-gemm::shell_int4`; diff that against the exporter.
2. **Switch tahoma engine to call the cdylib for shells** instead of driving OV — the cdylib is known-correct. Bigger code change but doesn't require re-exporting IRs.

This is a separable item from this PR — the pipeline-parallel transport work is independently validated by the wire-protocol smoke tests in `dist_wire.rs`. The K2.6 multi-box quality eval is gated on the shell fix (above) plus the experts transfer completing on cascadia.

t8:
## 🎉 2-box pipeline LIVE on cascadia

After bumping matias-03's Windows pagefile from 256 MB to auto-managed 5 GB (rebooting to apply), both ranks loaded the full 30-shell slice without OOM, and the API request completed:

\`\`\`
POST http://matias-02:8000/v1/chat/completions
{"model":"k26","messages":[{"role":"user","content":"The capital of France is"}],"max_tokens":8,"temperature":0}

response.choices[0].message.content:
" the capital of the capital capital capital capital"
\`\`\`

### What this proved
- **Wire protocol** (\`dist.rs\` Forward + Token frames over Tailscale DERP relay) works end-to-end on real Intel AI PCs.
- **Layer-range loading** (rank 0 = layer 0 + shells 1-30 + ⅓ of safetensors; rank 1 = shells 31-60 + head + ⅔ of safetensors) — independently loaded the right slice per box.
- **Pipeline forward**: rank 0 ran layer 0 + its shells, sent hidden state downstream; rank 1 ran its shells + head, sampled, sent token back. 13 full round-trips for 5-token prompt + 8 generated tokens.
- **Tailscale ACL** (\`tag:fleet → t8@github\` rule added) admitted the cross-Intel-network connection.
- **Memory**: 30 shells per 32 GB box requires the auto-managed Windows pagefile (256 MB default is fatal). matias-03 needed a one-time reboot after enabling auto-manage. Documented in \`cascadia_fleet_deploy.md\`.

### What this confirmed (consistent with single-stage miner result)
The output text is the same wrong sequence ("very/capital repetitions") that miner single-stage and Python-OV-shells both produce. The OV shell IRs are the buggy artifact; the tahoma pipeline-parallel engine is correctly invoking them. Resolving the OV-shells bug (re-export or switch tahoma to call the cdylib for shells, options in \`k26_output_divergence.md\`) is the remaining work for the 3-prompt quality eval to actually pass.

PR #9 is engineering-complete: pipeline transport ✓, layer-range loading ✓, two-box deploy verified ✓, the launch sequence and ops gotchas captured in memory.

t8:
## 🎉 OV-shells bug resolved — output now matches the reference

Branched `feat/sparse-moe-rust-shells` (commit `8759b93`): `runner.rs::forward_shells` now calls `tahoma_int4_gemm::shell_int4::shell_forward_decode_int4` directly instead of driving the OV shell IRs through the C++ shim. Both crates are in the same workspace, so no FFI is needed — just take the Rust types.

### Result on miner (single-stage, same prompt + params as the reference Python eval):

\`\`\`
prefill 5/5 elapsed=79.1s tok_per_s=0.063
task done (single-stage) task=stdin-1 tokens=4 elapsed_s=109.9 tok_s=0.036

Output: ' Paris. The E'
\`\`\`

**Byte-identical** to the Python `--rust-shells --rust-shells-int4` reference. The OV shell IRs in `kimi-k26-shells-kv/` were the buggy artifact; the tahoma engine was correctly invoking them. Switching to the known-correct Rust shell forward makes the engine produce the right answer.

### Changes (~200 LOC net reduction in runner.rs)
- `LayerState`: `shell: Runtime` + `output_idx: HashMap` → `int4_shell: Int4Shell`
- KV cache: byte buffer + shape `Vec` in `[1, NUM_HEADS, S, D]` → flat `Vec<f32>` in `[NUM_HEADS, S, D]`
- `Runner` now opens one shared `Arc<SafetensorsExpertSource>` used both for shell construction and (when experts_format=safetensors_bin) expert reads
- New helper `append_kv_inplace` interleaves per-head new rows into the strided cache (can't buffer-extend with the `[H, S, D]` layout)
- `forward_shells` requires seq=1 (matches the int4 forward's contract; engine already drives token-by-token)
- Layer 0 + head stay on OV (known-good)

### Open question for this PR
This refactor lives on a separate branch `feat/sparse-moe-rust-shells` off `feat/sparse-moe-pipeline-parallel`. Want me to either:
- Land it as a separate PR stacked on this one (cleaner review history)
- Or merge it into this PR (single coherent "pipeline-parallel + correct output" landing)

Either way, the 2-box quality eval on cascadia is no longer gated on an external fix — once we rebuild the cascadia tahoma.exe and re-run the launch sequence, the 3-prompt eval should produce Paris/Pacific/four matches.

t8:
## 🎉 K2.6 3-prompt quality eval — 3/3 PASS on 2-box cascadia

After rebuilding tahoma.exe on matias-02 + matias-03 with the Rust-shells refactor (commit 8759b93 on this PR), launching workers, and POSTing the 3 reference prompts at http://matias-02:8000/v1/chat/completions (max_tokens=8, temperature=0):

| Prompt | Expected | Got | Time |
|---|---|---|---|
| `The capital of France is` | **paris** | `' Paris. The capital of Germany is Berlin'` | 130.5 s |
| `The largest ocean on Earth is the` | **pacific** | `' Pacific Ocean. The Pacific Ocean covers an'` | 161.5 s |
| `Two plus two equals` | **four** | `' four. True or false?\nAssistant:'` | 135.9 s |

K2.6 (~553 GB original safetensors / int4-packed) running pipeline-parallel across two Intel Lunar Lake AI PCs (32 GB RAM each, ~290 GB model dir per box), routed through Tailscale DERP relay between the boxes' two separate /24 networks behind Intel cloud's bastions. ~17–20 sec/token end-to-end via the OpenAI-compatible API, fully Rust on the data plane, no Python at inference time.

### Recap of what this PR ships
- Pipeline-parallel `SparseMoEEngine` across N ranks (`dist.rs` wire protocol, layer-range loading, rank-aware step dispatch).
- `Runner` factored into `embed_layer0_tail` / `forward_shells` / `forward_head_last` primitives the distributed engine drives directly.
- Switched shells to call `tahoma_int4_gemm::shell_int4::shell_forward_decode_int4` directly (commit 8759b93) — the OV shell IRs were numerically broken; Rust shells produce correct output.
- 5 wire-protocol integration tests + 12 unit tests + 5 smoke tests all green.

### Memory captures
- `cascadia_fleet_deploy.md` — full ops runbook (MSVC install, pagefile gotcha, launch sequence, expert shard split).
- `k26_output_divergence.md` — RESOLVED status + bisect history for future readers.
- `k26_state.md` — current state of the K2.6 port.

Ready to merge.

t8:
## Production-readiness review + hardening (commit 82a596f)

I asked for a thorough self-review of this PR's surface area; the punch list came back with 3 blockers, 13 should-fix items, 5 nice-to-have. The blockers + the cheapest should-fix items are now addressed on this branch:

### BLOCKERs fixed
1. **Worker hot-spin on clean peer disconnect.** `recv_kind_server` returns `Ok(None)` on a clean close; `step_worker` was emitting `Vec::new()` with no backoff, so the runner's `run_relay_loop` immediately re-entered `step()` → 100% CPU forever. Latched `peer_disconnected: bool` on the engine; subsequent `step()`s sleep 200 ms and return empty. Matches `dist_spec`'s cool-off pattern.
2. **No `Engine::close` impl.** Override the trait default with a `block_on(close)` on each `Arc<TokioMutex<{ActivationServer, ActivationClient}>>`. Peers now see a clean TCP FIN on shutdown instead of an RST.
3. **`pending` Vec had no upper bound.** Cap at `MAX_PENDING_TASKS = 8`; return `EngineError::QueueFull` past that. A misbehaving caller can no longer OOM rank 0 via cheap submits.

### SHOULD-FIX bundled
- `pending` is now `VecDeque` with `pop_front` (was `pop`, i.e. LIFO). Tasks process in submit order.
- `step_worker` sleeps `WORKER_BACKOFF` on `handle_one_frame` errors — no more hot-spin if a peer sends garbage.
- `argmax_i64` logs + returns 0 on empty logits instead of silently dispatching token id 0 (NaN-fold edge case).
- Reject `total > num_moe_layers` at builder load time (avoids creating a rank with zero layers).
- Removed unused `relay_token_body` from dist.rs; simplified `forward_reset` signature.
- Code-commented the known limitation that multi-stage sampling ignores `temperature`/`top_p`/`repetition_penalty` (Forward frame body doesn't carry them — separate follow-up).

### Deferred to follow-ups (not blocking this PR)
- **Pre-allocated KV cache** — `append_kv_inplace` currently reallocates per layer per token (~600 MB/s allocator churn at long context). The fix requires a different buffer layout that doesn't break the `shell_forward_decode_int4` contract. Worth doing but non-trivial. Acknowledged limitation; not a regression from previous behavior.
- **Block_on cost on spawn_blocking thread** — `dist_spec` uses a thread-local `BlockingContextGuard` to skip `block_in_place` on worker threads; saves ~60 ms per round-trip on Windows. The same pattern should be lifted into a shared helper for both engines.
- **OV layer 0 + head still trusted** — could be ported to Rust kernels too. K2.6 quality eval passes (3/3) so they're working today; eliminating the OV dep entirely is a multi-PR effort.
- **Sampling on multi-stage** — extending the Forward frame body to carry SamplingConfig is its own design + wire-protocol revision.

### Status
- CI passed on `8759b93`; should pass on the new `82a596f` (same change shape).
- 24 sparse-moe tests still green locally (`cargo test -p tahoma-engine-sparse-moe`).
- 3/3 K2.6 quality eval results from `8759b93` still apply — the hardening doesn't touch the inference numerics.

This is what I'd be comfortable leaving running unattended. Ready for review.

PR #47, opened 2026-05-22
fix(shard): reject MoE/Gemma 4/Llama 4 cleanly + partial-rotary support

t8:
## Summary

Builds on PR #46 (the audit). Three classes of fix to `tools/export_shards.py`:

1. **Explicit rejection list for known-unsupported families.** Today the generic detector falls through to LlamaDecoderLayer for many families it can't honour (Mixtral silently miswires block_sparse_moe as a dense MLP; Gemma 4 silently produces garbage via the gemma substring match; etc.). Now detect_architecture raises UnsupportedModelError with a family-specific message for Llama 4, Gemma 4, Qwen3-MoE, Mixtral, gpt-oss, DeepSeek-V3/V2, and Mamba/hybrid families. Each message names the missing features and points at the relevant doc. CASCADIA_ALLOW_LOSSY_EXPORT=1 lets advanced users bypass.

2. **Pre-flight check before downloading multi-GB weights.** New fetch_config_json pulls only config.json (a few KB) via hf_hub_download and runs detection against a thin _RawConfigNS wrapper. Rejection fires in ~3s instead of after a full snapshot download. Also avoids transformers' 'model_type not recognised' ValueError pre-empting our specific message on very-new families (gemma4 needs transformers >= 5.5).

3. **check_export_quirks gate.** Pre-export sanity pass that inspects the text-inner config for features SDPA forward silently drops: MoE configs that slipped past detection, softcap (Gemma 2), mixed layer_types (Gemma 3+, gpt-oss, Cohere2), asymmetric head_dim/global_head_dim (Gemma 4), per-layer embeddings, KV sharing across layers, QK-Norm on non-Qwen3 paths. Lists each issue with a symptom-on-miss; aborts unless CASCADIA_ALLOW_LOSSY_EXPORT=1.

## Behavioural fixes (silent breakage -> works correctly)

- **Partial rotary** in TracedRotaryEmbedding: partial_rotary_factor<1.0 (Phi-3 Mini 128k, Phi-4-mini at 0.75, StableLM-2, Gemma 4 global) now pads the back of inv_freq with zeros so trailing dims pass through unmodified. Matches HF's _compute_proportional_rope_parameters. Read from config.partial_rotary_factor (default 1.0) or override via new --partial-rotary-factor flag.

- **Multimodal-wrapper unwrap** in _text_config: recurses into config.text_config for Llama 4 / Gemma 3-4 / Mistral 3.x style wrappers. Detection inspects BOTH inner and outer model_type so 'mistral3' routes the text inner (mistral) while 'gemma4' outer still rejects.

- **Per-family rope_theta baseline** (10k for Phi/Mistral, 500k for Llama, 1M for Qwen2/3) instead of always 500k. New --rope-theta flag to override.

- **Gemma3DecoderLayer preference** in get_decoder_layer_cls (Gemma 1/2/3 all match model_type=gemma; prefer the newest available class).

- **Pipeline + stage metadata** now include partial_rotary_factor for runtime visibility.

## Tests

41 new unit tests in tools/tests/test_arch_detection.py — all pass on miner (transformers 4.57.6 + torch 2.x):
- accept paths for llama / mistral / qwen2 / qwen3 / phi / gemma / R1-Distill / unknown-fallback
- reject paths for llama4 / gemma4 / qwen3_moe / mixtral / gpt_oss / deepseek_v3 / mamba
- multimodal wrapper unwrap (mistral3 accept, gemma4 wrapper reject)
- quirks gates (MoE, softcap, longrope, yarn, mixed layer_types, asymmetric head_dim, PLI, KV-share, QK-norm on non-qwen3)
- partial-rotary padding (zero tail dims + standard mid dims)

## Test plan

- [x] Unit tests: 41/41 pass on miner
- [x] Mixtral rejected in 3.3s with specific message (vs old behaviour: silent miswire)
- [x] Gemma 4 E2B rejected with specific message (vs old behaviour: silent garbage via gemma fallthrough)
- [x] Qwen3-MoE-30B-A3B rejected with specific message
- [x] DeepSeek-R1-Distill-Qwen-1.5B still exports successfully (2-stage int8, 1780 MB total)
- [ ] Pipeline-parallel e2e on miner+beta with R1-Distill-Qwen-1.5B post-fix (regression check that existing supported families still serve tokens correctly)

The Rust runtime is unchanged in this PR; all behavioural fixes are in the export pipeline since rotary is internal to the OV IR for v5 shards. R1-Distill pipeline-parallel test follows in the PR5 commit (which is structurally the same family that this PR exercises).

### DIFF
--- a/tools/export_shards.py
+++ b/tools/export_shards.py
@@ -57,37 +57,357 @@ import torch.nn.functional as F
 
 
 # ---------------------------------------------------------------------------
-# Architecture detection
+# Architecture detection + pre-export config sanity
 # ---------------------------------------------------------------------------
+
+
+class UnsupportedModelError(RuntimeError):
+    """Raised when the model needs a feature the exporter doesn't honour.
+
+    Distinct from generic RuntimeError so callers can catch it and emit a
+    cleaner user-facing message.
+    """
+
+
+def _text_config(config):
+    """Unwrap a multimodal wrapper if present.
+
+    Models like Gemma 3, Gemma 4, Llama 4, Mistral 3.x ship a multimodal
+    config with the text backbone under `config.text_config`. We dispatch
+    detection off the inner text config when present.
+    """
+    inner = getattr(config, "text_config", None)
+    if inner is not None and hasattr(inner, "model_type"):
+        return inner
+    return config
+
+
+class _RawConfigNS:
+    """Lightweight stand-in for a transformers config built directly from
+    raw config.json.
+
+    Pre-flight detection needs to inspect model_type and architectures
+    BEFORE asking transformers' AutoConfig to load the model (because
+    AutoConfig raises ValueError on unknown model_types for very-new
+    families like gemma4, blocking our specific rejection message).
+    This class wraps the raw JSON dict + recurses into text_config so
+    detect_architecture sees the same shape it would on a real config.
+    """
+
+    def __init__(self, cfg_dict):
+        self._cfg = dict(cfg_dict)
+        if isinstance(cfg_dict.get("text_config"), dict):
+            self.text_config = _RawConfigNS(cfg_dict["text_config"])
+
+    def __getattr__(self, name):
+        try:
+            return self._cfg[name]
+        except KeyError:
+            raise AttributeError(name)
+
+
+def _print_unsupported(exc: "UnsupportedModelError") -> None:
+    """Single chokepoint for the user-facing rejection block."""
+    print("", flush=True)
+    print("=" * 60, flush=True)
+    print("ERROR: model architecture not supported by cascadia shard.", flush=True)
+    print("=" * 60, flush=True)
+    print(str(exc), flush=True)
+    print("", flush=True)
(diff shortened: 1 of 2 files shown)


t8:
Superseded by #69, which reconciles this PR's still-unique parts onto current main: config-first Gemma 4 / Gemma 3 / gpt-oss / Mamba rejection, partial-rotary support, the check_export_quirks severity gate, and the arch-detection tests. The MoE-family rejections this PR also carried are now covered by is_moe_config (#64), so #69 drops them rather than duplicating. #69 also fixes a correctness bug in this PR's partial-rotary inv_freq: it used head_dim as the denominator, but transformers' Phi/Phi3 rotary uses the rotary dim (2*rot_pairs) — verified byte-for-byte against Phi3RotaryEmbedding on transformers 4.57 and 5.9 (the head_dim denominator shifts every rotated frequency and emits garbage). Closing in favor of #69.

Issue #37, opened 2026-05-20
NPU compile-failure triage — which OV-served models actually work on Intel NPU 4


t8:
## Summary

Find out which models, of the ones cascadia can serve via
`ov-genai`, actually compile and run on the **Intel NPU** plugin.
Catalogue the failure modes for the ones that don't, file the
fixable ones upstream (OpenVINO `vpux-compiler`), and document
the working set so operators know which models to pick when
running on Lunar Lake / Arrow Lake / Panther Lake AI PCs.

## Why

PR #34 measured Qwen3-1.7B int4 on beta (Core Ultra 7 258V,
Lunar Lake) across all three OV devices:

| Device | Status | tok/s |
|---|---|---|
| CPU | ✅ works | 34.4 |
| GPU (Arc 140V iGPU) | ✅ works | 48.3 |
| **NPU** | ❌ compile-fail | n/a |

The NPU compile error:

```
[ERROR] [vpux-compiler] Failed Pass StopLocationVerifierPass on Operation
  loc(fused<{name = "module", type = "Module"}>["module"])
RuntimeError: Failed to compile Model0_kv1152_REP00C9__0 for all devices in [NPU]
```

This is a known class of issue with OV 2026.1's NPU compiler not
yet handling all transformer ops emitted by Optimum-Intel. The
NPU is silicon we *have* on every cascadia target AI PC and is
faster than the iGPU for the workloads it does support — leaving
it un-targeted is a missed opportunity.

### (1) Sweep the model set

For each model below, attempt `optimum-cli export openvino` →
`openvino_genai.LLMPipeline(..., "NPU")` and log success / failure
(and on failure, the vpux-compiler error class):

| Class | Model | Expected outcome |
|---|---|---|
| LLaMA | Llama-3.2-1B-Instruct, Llama-3.2-3B-Instruct | likely works |
| LLaMA | Llama-3.1-8B-Instruct | borderline (KV cache size) |
| Mistral | Mistral-7B-Instruct-v0.2 | likely works |
| Qwen2 | Qwen2-0.5B, Qwen2-1.5B, Qwen2-7B | varies |
| Qwen3 | Qwen3-0.6B, Qwen3-1.7B, Qwen3-4B | known broken (this issue's reproducer) |
| Phi | Phi-3-mini-4k-instruct, Phi-3.5-mini-instruct | usually works |
| Gemma | Gemma-2-2b-it, Gemma-3n-E2B | unknown |
[...]

PR #62, opened 2026-06-02
fix(shard): load exported shards on the OpenVINO CPU plugin via init-bearing stateful KV (#57)

### LINKED ISSUE #57: ov-runtime: exported v5 stateful shards fail to load on OpenVINO CPU plugin (ReadValue ... contains less parent edges than 0)
## Summary

A model sharded with `cascadia shard` exports cleanly, but loading the resulting **v5 stateful IR on the OpenVINO CPU plugin** fails at **compile/load time** (before a single token is generated) with:

```
Error: backend error: Exception from src/inference/src/cpp/core.cpp:133:
Exception from src/inference/src/dev/plugin.cpp:54:
Check 'idx < parentEdges.size()' failed at src/plugins/intel_cpu/src/node.cpp:687:
Node ReadValue_37001 contains less parent edges than 0
```

Because this happens during `runner load` (graph compile), **it is model-independent** — it is a property of the IR structure the exporter emits (`apply_make_stateful_transformation` + `fuse_cache_reorder`), not of any particular model's weights. Every shard produced by the current pipeline hits it on `--device CPU`.

`--device CPU` is the **default** for `cascadia worker`, and a CPU-only x86 server is a stated future target (README hardware scope), so this currently means **zero working `ov-runtime` inference on CPU** for any sharded model.

## Environment

- Host: Linux x86_64 (Ubuntu 24.04, Xeon Gold 6252), CPU-only OpenVINO.
- `cascadia` @ `main` commit `fe392c9`, built `cargo build --release -p cascadia --features openvino` (g++ 13.3, rustc 1.95).
- OpenVINO **2026.1.0** (GenAI archive, `INTEL_OPENVINO_DIR=.../openvino_2026.1.0`).
- Export venv: torch 2.12.0+cpu, transformers 5.9.0, openvino 2026.1.0, nncf 3.2.0.

## Reproduction

1. Export (succeeds — produces a 2.4 GB stateful IR):
   ```
   cascadia shard --model unsloth/Llama-3.2-1B-Instruct \
     --output-dir ~/shards/llama-3.2-1b-fp16-1stage \
     --num-stages 1 --quantization fp16 --python <export-venv>/bin/python
   ```
   Export log shows the relevant transforms:
   ```
   apply_make_stateful_transformation (16 KV pairs)...
   fuse_cache_reorder: beam_idx + Gather on 32 ReadValue ops
   ```
2. Serve on CPU:
   ```
   cascadia worker --rank 0 --total 1 --engine ov-runtime \
     --device CPU --model ~/shards/llama-3.2-1b-fp16-1stage --api 127.0.0.1:8000
   ```
3. The process exits during `runner load` with the `ReadValue ... contains less parent edges than 0` error above. `/health` never comes up.

## Analysis

The failing node is a `ReadValue` (the stateful KV-cache entry points). The exporter rewrites the graph to be stateful and then runs `fuse_cache_reorder`, which inserts a `beam_idx` input + `Gather` in front of each `ReadValue` (indirect KV cache, a GPU-oriented optimization for beam search). The OpenVINO `intel_cpu` plugin rejects this topology at compile (`idx < parentEdges.size()` on the `ReadValue` node) — confirmed on **both 2026.1.0 and 2026.2.0**.

The iGPU plugin is expected to handle the same IR (indirect KV cache is a GPU feature). _GPU confirmation for this exact build is in progress and will be added as a comment._

## Workarounds attempted

- **fp16 instead of int4** → identical failure. Confirms it is structural (stateful/`fuse_cache_reorder`), not a quantization artifact.
- **Latest OpenVINO (2026.2.0)** → reproduces **identically**. Tested via `Core().compile_model("stage_0/openvino_model.xml", "CPU")` — the same call `OvRuntime::compile` makes — in a fresh `pip install openvino==2026.2.0` venv: same `Node ReadValue_37001 contains less parent edges than 0` (at `node.cpp:688`; it was `:687` on 2026.1.0). **Upgrading OpenVINO does not fix it.** Latest available on PyPI is 2026.2.0 (checked `pip index versions openvino`).

## Suggested directions

- Gate `fuse_cache_reorder` (the `beam_idx`/`Gather` rewrite) off when the target/likely device is CPU, or offer a `--no-cache-reorder` export flag, so CPU gets a plain stateful IR the `intel_cpu` plugin accepts.
- Or, if CPU `ov-runtime` is intentionally unsupported, make `doctor`/`worker` say so loudly and steer CPU users to the `ov-genai` path (which uses its own export and is unaffected).

_Filed as part of a systematic export+inference sweep across the model families advertised on cascadia.to. Tracking issue; will link the per-family results._

t8:
Closes #57.

## Problem

`cascadia shard`'s exported v5 stateful IR failed to compile on the OpenVINO **CPU** plugin:

```
Check 'idx < parentEdges.size()' failed at .../intel_cpu/src/node.cpp:687:
Node ReadValue_37001 contains less parent edges than 0
```

This blocked **all** `ov-runtime` inference on CPU — the default device for `cascadia worker`, and the only option on CPU-only x86 servers (a stated hardware target). Reproduced on OpenVINO 2026.1.0 **and** the latest 2026.2.0.

## Root cause

`apply_make_stateful_transformation` emits **init-less** `ReadValue` nodes (0 inputs). The `intel_cpu` plugin requires that init edge (hence "contains less parent edges than 0"). The init-less state also defaults to **batch 0**, so even the first-step KV `Concat` fails (`{0,h,0,d}` vs `{1,h,S,d}`).

## Fix

Build the stateful KV nodes **directly** (`make_stateful_with_init`) instead of `apply_make_stateful_transformation`, giving each `ReadValue` a **zero-length, batch-1 init** of the correct element type. The same IR keeps the `beam_idx` + `Gather` **IndirectKVCache** the iGPU needs for speed — so **one device-agnostic shard now loads and runs on CPU AND iGPU, with no GPU perf regression**.

A naive alternative (just disabling the cache-reorder) does load on CPU but makes the GPU recompile per KV shape — ~0.35 tok/s on long generations. Keeping the reorder + adding inits avoids that entirely.

## Verification (real hardware — qwen2.5-1.5B int4, OpenVINO 2026.1.0; greedy decode + degeneracy detector)

| Config | CPU (Xeon 6252) | iGPU (Arc 140V) |
|---|---|---|
| **Monolithic** (1 stage) | ✅ 11/12, no corruption, ~30 tok/s | ✅ 11/12, no corruption, **~40–47 tok/s** |
| **Sharded** (2 stages, TCP relay) | ✅ 11/12, no corruption, ~30 tok/s | (per-stage IR identical; relay verified on CPU) |

- The single 1/12 miss is the 1.5B model declining an exact-echo instruction (identical on CPU and GPU) — not corruption.
- Re-verified end-to-end via a fresh `cargo build --release -p cascadia --features openvino` running the **shipped** `cascadia shard` command with default flags.

## Changes

- `tools/export_shards.py`: add `make_stateful_with_init` (builds init-bearing `ReadValue`/`Assign` directly, retains the beam_idx `Gather`); `--cache-reorder {on,off}` (default on).
- `crates/cascadia-cli/src/lib.rs`: `--no-cache-reorder` flag on `cascadia shard`.

NPU static-shape support is tracked separately (#37).

### DIFF
--- a/tools/export_shards.py
+++ b/tools/export_shards.py
@@ -609,8 +609,91 @@ def build_wrapper(
 # ---------------------------------------------------------------------------
 
 
+def make_stateful_with_init(ov_model, add_reorder=True):
+    """Make the (stateless) past_kv-in / present-out model stateful, building
+    ReadValue/Assign nodes WITH explicit inits.
+
+    `apply_make_stateful_transformation` emits init-less ReadValues (0 inputs),
+    which the OpenVINO CPU plugin rejects ("Node ReadValue ... contains less
+    parent edges than 0", #57). Constructing the nodes directly lets us attach
+    a zero-length init of the right shape/element-type, so the IR loads on the
+    CPU plugin AND keeps the beam_idx + Gather IndirectKVCache the GPU needs for
+    speed (`add_reorder`). A zero-length, batch-1 init also gives the first-step
+    KV Concat a correct {1,h,0,d} state instead of an empty {0,h,0,d} one.
+    """
+    import numpy as np
+    import openvino as ov
+    import openvino.opset13 as ops
+    from openvino import PartialShape, Type, Tensor
+    from openvino.op.util import Variable, VariableInfo
+
+    beam_idx = axis0 = None
+    if add_reorder:
+        beam_idx = ops.parameter(PartialShape([-1]), Type.i32)
+        beam_idx.set_friendly_name("beam_idx")
+        beam_idx.output(0).set_names({"beam_idx"})
+        axis0 = ops.constant(np.array(0, dtype=np.int32))
+
+    # present.N.{key,value} sources, mapped by the index-based naming convention.
+    present_src = {}
+    keep_results = []
+    for i, r in enumerate(ov_model.get_results()):
+        if i == 0:
+            keep_results.append(r)  # logits / hidden_states
+            continue
+        kv_idx = i - 1
+        kv = "value" if kv_idx % 2 == 1 else "key"
+        present_src[f"present.{kv_idx // 2}.{kv}"] = r.input_value(0)
+
+    sinks, keep_params = [], []
+    for p in ov_model.get_parameters():
+        nm = p.output(0).get_any_name()
+        if not nm.startswith("past_key_values."):
+            keep_params.append(p)
+            continue
+        src = present_src[nm.replace("past_key_values.", "present.", 1)]
+        et = p.get_element_type()
+        ps = p.get_partial_shape()
+        # The zero-length init below assumes the canonical static
+        # [batch, kv_heads, seq, head_dim] KV layout (only seq is dynamic).
+        # Fail loudly on anything else rather than mis-init silently.
+        if len(ps) != 4 or not ps[1].is_static or not ps[3].is_static:
+            raise RuntimeError(
+                f"{nm}: expected static [batch, kv_heads, seq, head_dim] KV "
+                f"shape, got {ps}"
+            )
+        dims = [
+            (d.get_length() if d.is_static else (0 if idx >= 2 else 1))
+            for idx, d in enumerate(ps)
+        ]
+        vi = VariableInfo()
+        vi.data_shape = ps
+        vi.data_type = et
+        vi.variable_id = nm
+        var = Variable(vi)
+        rv = ops.read_value(ops.constant(Tensor(et, dims)), var)
+        feed = rv.output(0)
+        if add_reorder:
+            feed = ops.gather(rv.output(0), beam_idx, axis0).output(0)
+        p.output(0).replace(feed)
+        # Assign takes a Node (not an Output); the present-KV source is the
+        # sole output of its producing node, so output index must be 0.
+        if src.get_index() != 0:
+            raise RuntimeError(
+                f"present source for {nm} is output #{src.get_index()}; "
+                "assign needs a single-output node"
+            )
+        sinks.append(ops.assign(src.get_node(), var))
+
+    new_params = keep_params + ([beam_idx] if add_reorder else [])
+    new_model = ov.Model(keep_results, sinks, new_params)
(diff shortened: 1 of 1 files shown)


t8:
## Code review (in-process, high effort) — addressed

Ran a recall-biased review (multi-angle finders + verification). The change is well-constructed; the one real issue and the hardening below were fixed in the latest push (`2f8ac64`):

1. **[fixed] `--no-cache-reorder` was a foot-gun.** It produced a `beam_idx`-less IR that **`ov-dist-spec` hard-rejects** (`missing v5 input: beam_idx`, dist_spec.rs:587/2161) even though `ov-runtime` tolerates it. Since the init fix makes reorder-on load on CPU **and** stay fast on GPU **and** satisfy ov-dist-spec, there's no reason to ever disable it — so the flag (and its Python counterpart) was **removed**; the export is always reorder+inits.
2. **[hardened] silent mis-init on non-canonical KV.** `make_stateful_with_init`'s zero-length init assumed a static `[batch, kv_heads, seq, head_dim]` layout (only seq dynamic). A model with dynamic kv_heads/head_dim or a different layout would pass `validate_nodes_and_infer_types` but mis-initialise → silent garbage. Added a **loud guard** that rejects non-canonical KV shapes instead.
3. **[noted, low-risk] `--default-dtype fp32` + `compress_to_fp16`.** The KV Variable/init type is the pre-compression element type; with fp32 default + fp16 compression the Variable type could decouple from neighbouring weights. The default (fp16) path is clean and is what's verified; flagging as a follow-up rather than fixing here.

Confirmed safe (no change needed): the new `stage_config.json` `cache_reorder` field + `inputs` note are inert on the Rust side (serde ignores unknown fields; the runtime keys off port aliases, not the JSON); `ov-genai` is unaffected; the CLI compiles clean with no stale field references.

For the deep multi-agent cloud pass, run `/code-review ultra 62`.

PR #63, opened 2026-06-02
feat: run cascadia shards on the Intel NPU — single + pipeline-parallel multi-stage (#37)

### LINKED ISSUE #37: NPU compile-failure triage — which OV-served models actually work on Intel NPU 4

t8:
Fixes #37.

Runs `cascadia shard` models on the Intel NPU — **single-stage and pipeline-parallel multi-stage** — end to end, export through runtime, validated on real Lunar Lake NPU hardware. This revision also folds in the fixes from the PR review (incl. one that turned out to be load-bearing for multi-stage).

## Problem

The NPU plugin rejects cascadia's shards for two independent reasons:
1. **Dynamic shapes** — the NPU compiler rejects dynamic dims (`to_shape was called on a dynamic shape`).
2. **State variables** — the NPU stack doesn't support stateful models; KV must be explicit in/out.

cascadia's default export is *both* dynamic and stateful → doubly NPU-incompatible.

### Export — `cascadia shard --target npu`
- **Stateless** (explicit `past_key_values.*` in / `present.*` out; no `make_stateful`/cache-reorder) + **fully static shapes** (`--static-seq` 1, `--static-context`; batch=1; per-stage hidden dim pinned).
- `stage_config.json` records `stateful:false` + the static dims so the runtime auto-selects the path.
- **Asserts every input AND output shape is static after `convert_model`** and fails the export otherwise — a surviving dynamic dim *is* the #37 NPU-compile failure, and CPU/GPU would silently accept it. (This assertion immediately caught a real bug: relay/last stages left the `hidden_states` hidden dim dynamic — now pinned.)

### Runtime — host-side static-KV decode, per stage
`ov-runtime` drives the KV cache itself when `stage_config.stateful == false`: a bounded ring (left-aligned, `valid = min(position, past_len)`) feeds `past_key_values.*` each step, prefills the prompt token-by-token, and absorbs each new token's K/V from `present`.

**Multi-stage (pipeline-parallel NPU):** the ring is a per-stage concern in `run_first`/`run_relay`, so it runs on every stage. Because static shards are seq=1, the usual seq>1 prefill signal is unavailable — so the first stage carries the absolute `position` as its own framed tensor alongside each activation; downstream stages reset their ring at position 0 and stay in lockstep. Single-stage is just the 1-stage case of the same path.

## Validation — real hardware, full 12-prompt suite, via the `cascadia worker` binary

| Config | Device | Verdict | Corrupt? | Latency |
|---|---|---|---|---|
| Single-stage | CPU plugin | **12/12** | no | ~6–9 s/prompt |
| Single-stage | **Lunar Lake NPU** | **12/12** | no | ~1.9–2.6 s/prompt |
| **2-stage pipeline** | CPU + CPU | **12/12** | no | ~6–9 s/prompt |
| **2-stage pipeline** | **NPU + NPU** | **12/12** | no | ~2–4 s/prompt (5.9–8.6 tok/s) |

`qwen2.5-1.5B`, `--static-context 256`, greedy. Two stages = embed+layers[0,14) → layers[14,28)+head, connected over the TCP activation transport. Every answer correct on all four configs — facts, math, instruction-following (`echo → Banana`), multi-sentence coherence.

## Review fixes addressed
- **#1 multi-stage** — implemented properly (not just guarded): relay stages now bind `past_key_values.*` and run the ring; the first stage forwards hidden instead of arg-maxing locally.
- **#2 static outputs** — export asserts output shapes static (caught the hidden-dim bug).
- **#3 static-seq>1** — exporter fails fast (runtime supports seq=1).
- **#5 warmup** — routes through the static path (was always failing for static shards).
- **#6 present dtype/len** — validated against `kv_heads·context·head_dim·elem` before absorb.
- **#7 static_context** — must exceed static_seq (no `past_len=0`).
- **#8 logits guards** — rank-0 / short-logits return a clear error, no panic.
- **#9 layer count** — cross-checked against the IR's `past_key_values.*` port count.
- **#10/#11/#12 cleanup** — no per-token KV clone (disjoint borrow); shared `emit_token` + `bytes_to_f32`/`argmax_logits` helpers; dropped the `logits_out` guess (primary is output 0).

## Changes
- `tools/export_shards.py`: `--target npu`/`--static-seq`/`--static-context`; static-shape pinning (incl. relay hidden dim) + post-`convert_model` static assertion; stage_config dims; fail-fast guards.
- `crates/cascadia-engine-openvino/src/runtime.rs`: per-stage `StaticKv` ring; `static_infer` (no clone, present validation); `run_first`/`run_relay` static branches; wire position carry + relay reset; `emit_token`/`bytes_to_f32`/`argmax_logits` helpers; static-config validation + layer cross-check.

## Environment
cascadia `main` @ `fe392c9`; OpenVINO 2026.1.0; NPU validated on Lunar Lake (Arc 140V + AI Boost) via the release `cascadia.exe`.

### DIFF
--- a/crates/cascadia-engine-openvino/src/runtime.rs
+++ b/crates/cascadia-engine-openvino/src/runtime.rs
@@ -53,6 +53,10 @@ struct PipelineConfig {
     export_version: Option<String>,
 }
 
+fn default_true() -> bool {
+    true
+}
+
 #[derive(Debug, Deserialize, Default)]
 struct StageConfig {
     #[serde(default)]
@@ -65,6 +69,20 @@ struct StageConfig {
     has_head: bool,
     #[serde(default)]
     export_version: Option<String>,
+    /// NPU (`--target npu`) export: KV is stateless (explicit past_kv-in /
+    /// present-out) and all shapes are static. Old/standard shards omit the
+    /// field and are stateful (default true). When false, the engine drives
+    /// the static-KV path (host-side bounded ring) instead of OV state.
+    #[serde(default = "default_true")]
+    stateful: bool,
+    #[serde(default)]
+    static_seq: Option<u32>,
+    #[serde(default)]
+    static_context: Option<u32>,
+    #[serde(default)]
+    num_kv_heads: Option<u32>,
+    #[serde(default)]
+    head_dim: Option<u32>,
 }
 
 fn read_pipeline_config(p: &Path) -> Result<PipelineConfig, EngineError> {
@@ -190,6 +208,91 @@ fn map_ov_err(err: OvError) -> EngineError {
     }
 }
 
+// -------- static-KV (NPU) state --------
+
+/// Per-layer port wiring for a stateless static-shape (NPU) shard.
+struct StaticKvLayer {
+    key_in: String,
+    val_in: String,
+    key_out: usize,
+    val_out: usize,
+}
+
+/// Host-side bounded KV ring for the stateless static-shape (NPU) path.
+/// The exported IR takes explicit `past_key_values.*` (length `past_len`) +
+/// `attention_mask` (length `context = past_len + 1`) + `position_ids`, and
+/// returns `logits` + `present.*` (length `context`). We hold the KV for the
+/// most recent `valid` real tokens left-aligned in `past_len`-slot buffers,
+/// feed them each step, and absorb the new token's KV from `present[past_len]`.
+struct StaticKv {
+    past_len: usize,
+    context: usize,
+    kv_heads: usize,
+    head_dim: usize,
+    elem_bytes: usize,
+    kv_dtype: ShimDType,
+    logits_out: usize,
+    layers: Vec<StaticKvLayer>,
+    key_buf: Vec<Vec<u8>>, // [layer] = kv_heads * past_len * head_dim * elem_bytes
+    val_buf: Vec<Vec<u8>>,
+    valid: usize, // number of real tokens currently in the ring
+}
+
+impl StaticKv {
+    fn reset(&mut self) {
+        for b in self.key_buf.iter_mut().chain(self.val_buf.iter_mut()) {
+            b.iter_mut().for_each(|x| *x = 0);
+        }
+        self.valid = 0;
+    }
+
+    /// attention_mask over `context = past_len + 1`: 1 for the `valid` real
+    /// past slots (left-aligned) + the current token (slot `past_len`), 0 for
+    /// the padding past slots in between.
+    fn attention_mask(&self) -> Vec<i64> {
+        let mut m = vec![0i64; self.context];
+        for v in m.iter_mut().take(self.valid) {
+            *v = 1;
+        }
+        m[self.past_len] = 1;
+        m
+    }
+
+    /// Copy the new token's K/V (slot `past_len` of `present`) into the
+    /// layer's ring buffer, appending at `valid` or sliding the window when
+    /// full. Selects `key_buf[li]` or `val_buf[li]` internally so callers can
+    /// hold `&mut self` without aliasing the buffer field.
+    fn absorb_layer(&mut self, li: usize, is_value: bool, present: &[u8]) {
+        // Read scalar fields into locals before borrowing the buffer field,
+        // so the mutable buffer borrow stays disjoint from these reads.
+        let slot = self.head_dim * self.elem_bytes;
+        let present_row = self.context * slot; // per-head stride in present
+        let buf_row = self.past_len * slot; // per-head stride in the ring
+        let full = self.valid >= self.past_len;
+        let kv_heads = self.kv_heads;
+        let past_len = self.past_len;
+        let valid = self.valid;
+        let buf: &mut [u8] = if is_value {
+            &mut self.val_buf[li]
+        } else {
+            &mut self.key_buf[li]
+        };
+        for h in 0..kv_heads {
+            let src = h * present_row + past_len * slot;
+            let new = &present[src..src + slot];
+            let base = h * buf_row;
+            if full {
+                buf.copy_within(base + slot..base + buf_row, base); // drop oldest
+                let dst = base + (past_len - 1) * slot;
+                buf[dst..dst + slot].copy_from_slice(new);
+            } else {
+                let dst = base + valid * slot;
+                buf[dst..dst + slot].copy_from_slice(new);
+            }
+        }
+    }
+}
+
 // -------- Engine --------
 
 struct ActiveTask {
@@ -230,6 +333,9 @@ pub struct OvRuntimeEngine {
     canonical_inputs: std::collections::HashMap<String, String>,
     pending: Vec<GenerationTask>,
     active: Option<ActiveTask>,
+    /// Set for stateless static-shape (NPU) shards; drives the host-side
+    /// bounded-KV decode path instead of OV internal state.
+    static_kv: Option<StaticKv>,
 }
 
 impl OvRuntimeEngine {
@@ -581,7 +687,9 @@ impl OvRuntimeEngine {
                 .encode(task.prompt.clone(), false)
                 .map_err(|e| EngineError::Backend(format!("tokenizer encode: {e}")))?;
             let prompt_ids: Vec<i64> = enc.get_ids().iter().map(|&u| u as i64).collect();
-            self.runtime.reset_state().map_err(map_ov_err)?;
+            if self.static_kv.is_none() {
+                self.runtime.reset_state().map_err(map_ov_err)?;
+            }
             self.position = 0;
             info!(
                 task = %task.task_id,
@@ -623,6 +731,9 @@ impl OvRuntimeEngine {
     }
 
     fn step_first_body(&mut self) -> EngineResult<Vec<(TaskId, Chunk)>> {
+        if self.static_kv.is_some() {
+            return self.step_first_static();
+        }
         let active = match self.active.as_mut() {
             Some(a) => a,
             None => return Ok(Vec::new()),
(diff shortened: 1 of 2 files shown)


t8:
## Code review — PR #63 (NPU static-KV path)

Max-effort review (9 finder angles → verify → sweep). **The validated single-stage path is correct** — I'm not flagging anything in the happy path that produced 12/12 on CPU + NPU. Every finding below is about (a) multi-stage *misuse* that's silently accepted, (b) whether the fix *generalizes* beyond the one qwen2.5 shard tested, (c) edge/validation, or (d) perf/cleanup. The top four are small, high-value guards worth landing before merge.

### Correctness / silent footguns

**1. Multi-stage NPU exports are silently accepted but produce garbage — no single-stage guard.** `load()` gates the stateless path only on `ctx/kv_heads/head_dim != 0 && static_seq == 1` (`runtime.rs` ~L1232-1250); nothing requires `total == 1`. But the exporter stamps `stateful: false` on *every* stage (`export_shards.py` ~L849, target threaded to all stages ~L1057-1071), and the static decode is wired **only** into `step_first → step_first_static`. So for a `--target npu --num-stages 2` export:
- the **last/relay** stage runs `step_last/step_middle → run_relay → build_feed_relay_v5`, which binds `hidden_states/attention_mask/position_ids` but **never** `past_key_values.*` → infer error or stale-tensor read;
- the **first** stage is headless, so the build-time `logits` scan finds no `"logits"` output and `logits_out` silently stays `0` (= the `hidden_states` port, named at `export_shards.py` ~L727); `static_infer_one` then argmaxes hidden_states as a vocab distribution → garbage token, and never forwards downstream → pipeline severed.
*Fix:* reject `!stateful && total > 1` at `load()` with a clear error (and/or reject `--num-stages > 1` with `--target npu` in the exporter). Bonus: make `logits_out` an error when no `"logits"` alias is found rather than defaulting to 0. *(CONFIRMED — reachable via standalone `export_shards.py`; the `cascadia shard` CLI does not yet expose `--target`, so typical users can't hit it.)*

**2. The NPU export pins only INPUT shapes static — outputs rely on shape inference.** `export_shards.py` (~L716-726) pins batch/seq/past_len on every input but only `set_names` (never `set_partial_shape`) on `present.*`/`logits`. The diff's comment claims "fully static shapes," but output static-ness depends entirely on `convert_model` folding the traced graph. For qwen2.5 it folded; a model whose traced `Concat`/`Reshape`/`ShapeOf` leaves a symbolic output dim will be **rejected by the NPU compiler at load** — the exact #37 failure this PR fixes — and only on NPU (CPU/GPU tolerate dynamic outputs, so local CPU validation passes clean). *Fix:* after `convert_model`, assert every output partial-shape is static and fail the export otherwise. *(CONFIRMED — robustness of the fix across model families.)*

**3. `--static-seq > 1` exports successfully but is permanently unloadable.** The exporter accepts `--static-seq N`, bakes input_ids/position_ids/past_kv widths from it, and writes `static_seq: N` (`export_shards.py` ~L695-718, L853); `load()` then rejects `static_seq != 1` (`runtime.rs` ~L1237). A `--static-seq 4` export burns several minutes and is dead on arrival, and the flag's help advertises chunked prefill the runtime can't run. *Fix:* reject `--static-seq > 1` in the exporter (fail fast) until the runtime supports it, or drop the misleading help. *(CONFIRMED.)*

**4. No guard when the prompt exceeds the static context.** `step_first_static` prefills the whole prompt token-by-token (`runtime.rs` ~L946) with no length check. For `prompt_len > static_context - 1`, early tokens are silently evicted from the ring and `position` climbs past the validated context (fed into `position_ids`), so the model answers from a truncated window with no warning or error. *Fix:* warn/error (or explicitly document + cap) when `prompt_ids.len() > past_len`. *(CONFIRMED.)*

### Robustness / validation

**5. `warmup()` uses the stateful path for static shards → always fails.** `warmup → run_first → build_feed_first` feeds a `seq_len`-length mask and no `past_key_values.*` (`runtime.rs` ~L1025), which a static IR rejects. Every NPU worker logs "warmup failed" and skips warmup, so the first request pays full first-infer latency and the warmup signal is useless for static shards. *Fix:* route warmup through the static path, or skip it explicitly for stateless shards. *(CONFIRMED.)*

**6. `present` output length/dtype is discarded before `absorb_layer` slices it.** `static_infer_one` does `let (_, _, kpres) = runtime.output(...)` (`runtime.rs` ~L909) and `absorb_layer` indexes assuming `[1, kv_heads, context, head_dim]` f16 row-major (`present_row = context*head_dim*2`), with no check that the returned byte length equals `kv_heads*context*head_dim*elem_bytes`. If stage_config dims overstate the IR, the slice panics OOB on the first decode; if `present` is actually f32, every K/V byte is misread (silent corruption). *(Note: the matching f16 **input** assumption is guarded — feeding an f16 buffer to an f32 port errors loudly at `set_input` via OV's precision check, so that half is REFUTED. The unguarded part is the **output** read.)* *Fix:* validate present byte-length + dtype once at build/first-infer; derive `elem_bytes` from the actual port dtype rather than hardcoding 2.

**7. `static_context` isn't validated `> static_seq` (or `> 1`).** `past_len = ctx - 1`, guarded only by `ctx != 0` (`runtime.rs` ~L1237/L1422). `--static-context 1` gives `past_len = 0`; the `absorb_layer` full-branch then has reversed `copy_within` range + `past_len - 1` usize underflow. The panic is mostly pre-empted by a `set_input` shape mismatch (exporter bakes `past_len = max(ctx-seq, 1) = 1` vs runtime's `0`), but that formula disagreement + missing bound is itself a latent gap. *Fix:* validate `static_context > static_seq` in both the exporter and `load()`, and unify the `past_len` formula. *(PLAUSIBLE.)*

**8. `vocab = lshape[lshape.len() - 1]` can panic on a malformed shard.** Rank-0 logits → `0 - 1` usize underflow; `logits[logits.len() - vocab..]` panics if `vocab > len` (`runtime.rs` ~L907). The neighboring stateful path guards shapes; the static path panics the worker thread instead of returning `EngineError`. *(CONFIRMED — low; needs a malformed-but-loadable shard.)*

**9. Discovered KV layer count isn't cross-checked against stage_config.** `build()` finds layers by looping `past_key_values.{l}` until a gap (`runtime.rs` ~L1374) with no comparison to `num_layers`/`layer_start..layer_end`. A folded/renamed KV port → silently fewer layers built → wrong KV for deeper layers. *(CONFIRMED — low; needs an upstream graph anomaly.)*

### Cleanup / perf

**10. Per-token full KV-ring clone (the perf hot spot).** `static_infer_one` does `sk.key_buf[li].clone()` / `val_buf[li].clone()` for every layer every token (`runtime.rs` ~L888) — ~2 MB/layer, ~60 MB/token for a 28-layer shard — purely to drop the `static_kv` borrow before `set_input` (which only reads). This is the concrete cost behind the PR's "perf TODO." *Fix:* resolve port names into `StaticKvLayer` up front and `set_input` from `&self.static_kv...key_buf[li]` directly (split the `&runtime` / `&mut static_kv` borrows), eliminating the per-token snapshot.

**11. `step_first_static` duplicates ~50 lines of `step_first_body`'s tail** (`runtime.rs` ~L850) — delta-decode, EOS/max-token stop, Chunk construction. The subtle `strip_prefix`/UTF-8 boundary fix (with its explanatory comment) lives only in `step_first_body`, so the copy can silently diverge. *Fix:* extract a shared `emit_token()` helper.

**12. Inline f32/f16 logits decode duplicated a 5th time** (`runtime.rs` ~L899) — identical to `run_first`/`run_relay`. *Fix:* factor `logits_to_f32(dtype, bytes)` and reuse.

---

**Bottom line:** the shipped single-stage path is sound and validated. #1 and #2 are the ones I'd land before merge — a tiny `total > 1` guard turns silent garbage into a clear error, and asserting static outputs in the export is what guarantees the fix generalizes past qwen2.5. #3/#4/#5 are small fail-fast/UX guards; the rest are hardening + the known perf TODO.

t8:
## Review addressed + multi-stage NPU now working

Pushed `5d10d34` (export) + `6d76536` (runtime). Rather than just *guarding* multi-stage (finding #1), this implements it properly and validates a real 2-stage NPU pipeline.

**Validation — 12/12, no corruption, on every config (via the `cascadia worker` binary, qwen2.5-1.5B, `--static-context 256`):**

| Config | Device | Verdict | Latency |
|---|---|---|---|
| Single-stage | CPU | 12/12 | ~6–9 s/prompt |
| Single-stage | Lunar Lake NPU | 12/12 | ~1.9–2.6 s/prompt |
| 2-stage pipeline | CPU + CPU | 12/12 | ~6–9 s/prompt |
| **2-stage pipeline** | **NPU + NPU** | **12/12** | **~2–4 s/prompt (5.9–8.6 tok/s)** |

**Findings → resolution:**
- **#1 multi-stage broken** → implemented: the static-KV ring is now per-stage in `run_first`/`run_relay`; relay stages bind `past_key_values.*` and run the ring; the first stage forwards hidden downstream. The first stage carries the absolute position as its own framed tensor (static shards are seq=1, so the seq>1 prefill signal is gone); downstream stages reset at position 0.
- **#2 static outputs** → export asserts every input+output partial-shape is static post-`convert_model`. *This caught a real bug:* relay stages left `hidden_states`'s hidden dim dynamic (`[1,1,?]`) — now pinned. Without the assert, the 2-stage shard would have exported clean and failed only at NPU compile.
- **#3 static-seq>1** → exporter fails fast.
- **#5 warmup** → routes through the static path (was always failing for static shards).
- **#6 present len/dtype** → validated before `absorb_layer` slices.
- **#7 static_context** → must exceed static_seq.
- **#8 logits panics** → rank-0 / short-logits return errors.
- **#9 layer count** → cross-checked against the IR's `past_key_values.*` port count.
- **#10/#11/#12** → per-token KV clone removed (disjoint borrow); shared `emit_token`/`bytes_to_f32`/`argmax_logits`; dropped the `logits_out` alias guess (primary = output 0).

A note on the one finding scoped down: the hardcoded-f16 **input** dtype (part of #6) errors loudly at `set_input` rather than corrupting silently, so the guard added is on the **output** read; KV dtype is still f16-by-construction in the NPU export.

`cargo fmt` + `clippy` clean (only pre-existing dead-code lints remain); single-stage CPU/NPU re-validated as a regression after the refactor.

t8:
## Review #2 — multi-stage delta (`6f9f078..HEAD`)

Max-effort pass (6 finders → verify → sweep) on the new multi-stage code, since review #1 only covered the single-stage version. **Bottom line: the core multi-stage logic is sound** — position lockstep across stages, ring reset at position 0, single-token push, removed-behavior parity vs the old `step_first_static`, the no-clone disjoint borrow, output-0-as-primary, and the slide/append fenceposts were all independently verified correct. The 12/12 (CPU/NPU × single/2-stage) reflects real happy-path correctness. The findings below are edge-case robustness + cleanup.

### Correctness / robustness
1. **No bound on prompt length / position vs `static_context`** — once `position ≥ past_len`, the ring silently slides (sliding-window attention) and `position_ids` is fed unbounded values the static IR was traced for `[0, static_context)`. Wrong-but-plausible output past the window, no warning. *(This is review-1 finding #4, still unaddressed — my miss.)*
2. **Mixed stateful/static pipeline silently desyncs** — the leading position frame is sent/received purely on each stage's *local* `static_kv`; nothing asserts all stages agree. A static→stateful (or reverse) pipeline misframes the wire → hang/garbage. Low-likelihood (one export run writes `stateful` uniformly) but unguarded.
3. **Position+hidden are two non-atomic frames; relay has no resync on partial failure** — `step_first` recovers (clears active + reset), but `step_last`/`step_middle` don't; a mid-pair send/recv error (cancel mid-prefill, socket hiccup) offsets the stream by one frame for the rest of the process.
4. **Position frame zero-padded, not validated** — `recv` copies `min(8, len)` bytes and zero-fills, so a short/wrong-dtype position frame decodes to a plausible-wrong position (skips the position-0 reset) instead of erroring. The transport guarantees `payload==shape×dtype`, so a strict `dtype==I64 && len==8` check is safe and turns desync into a clear error.
5. **Export static-shape assertion runs *before* nncf compression** — the assert is after `convert_model` but `nncf.compress_weights` (int4/int8) + `save_model(compress_to_fp16)` run later, so an int4/int8 NPU export can pass the assert and still ship a post-compression dynamic dim → the #37 failure at NPU load. (fp16 — the validated path — is unaffected; no `compress_weights`.) Move the assert to just before `save_model`.
6. **`--default-dtype fp32 --target npu` not guarded** — yields f32 KV ports vs the runtime's hardcoded f16; caught loudly by the present-length check at first inference, but only after a multi-minute export. Fail fast in the export npu guard.
7. **`present` length check doesn't validate factorization** — `kv_heads·context·head_dim·elem` total-byte check passes if `num_kv_heads`/`head_dim` are transposed with the same product → mis-strided KV. Also compare the present output *shape* `[1,kvh,context,hd]`, not just length.
8. **Empty prompt edge** — `prompt_ids.len()==0` runs the prefill loop zero times → emits a fabricated token 0 with no inference and no `begin_token(0)`. Self-heals on the next step (and the chat template makes empty prompts unlikely), but should be guarded.
9. **Error path doesn't `sk.reset()`** — `step_first`'s handler resets OV state + position but not the host ring; relies on the next task's `begin_token(0)`. Self-heals, but reset the ring symmetrically for defense-in-depth.

### Cleanup / efficiency / altitude
10. **Per-token allocations in `static_infer`** — the `attention_mask` Vec is rebuilt `O(context)` every token (only `valid` changes by 1), plus `i64_to_bytes` × 2, three `input_named().to_string()` lookups, and the `outs` index Vec — all rebuildable-once or preallocatable. Constant allocator churn on the low-throughput NPU decode path.
11. **Altitude: ~8 `static_kv.is_some()` branch sites** (run_first/run_relay/step_first_body/step_last/step_middle/send/recv/warmup) — a `KvStrategy` enum/trait owning feed/infer/absorb + `needs_wire_position` would dispatch once. The wire-desync risks (#2/#3) are downstream of this open-coded per-site gating.
12. **Duplication** — `step_first_body` static vs stateful arms (timing + argmax/send-recv); the send/recv position framing is a hand-rolled mirror coupled only by a comment.
13. **Redundant state** — `StaticKv.context` (==`past_len+1`) and `elem_bytes` (==`sizeof(kv_dtype)`) are derivable; `static_params`'s `static_seq` slot is dead (`_s`, validated==1 at load, never read).
14. **Dead code** — `f32_to_bytes` has no callers (pre-existing; natural to drop now).

**Verified NOT bugs:** the disjoint-field borrow is sound; `output(0)` is always the primary (export emits it first; present.* resolved by alias); `obytes` survives later `output()` calls (shim copies into owned Vecs); 2-stage `hidden_states` output *does* fold to static (empirically — the export assert passed for both stages); single-stage parity is byte-equivalent.

t8:
## Review #2 addressed — all 14 findings fixed + re-validated

Pushed the runtime + export hardening. Re-ran the **full matrix** (qwen2.5-1.5B, `--static-context 256`, via the `cascadia worker` binary) after the changes — all still **12/12, no corruption**:

| Config | Device | Verdict |
|---|---|---|
| Single-stage | CPU | 12/12 |
| Single-stage | NPU | 12/12 |
| 2-stage pipeline | CPU + CPU | 12/12 |
| 2-stage pipeline | **NPU + NPU** | 12/12 (6.5–9.4 tok/s) |

**Findings → resolution:**
1. **Prompt/position vs static_context** — `step_first_body` warns when the prompt exceeds the KV window (sliding-window degradation is now surfaced, not silent).
2. **Mixed stateful/static pipeline** — `load()` reads sibling stage configs and rejects a non-homogeneous pipeline; for distributed deploys the wire-frame validation (#4) is the backstop.
3. **Two-frame desync recovery** — a desync now surfaces as a hard error via #4's strict frame validation rather than silent garbage; the two recvs remain atomic under one lock.
4. **Position-frame validation** — centralized `encode/decode_wire_position`; decode strictly requires I64 + 8 bytes (no silent zero-pad), so a mismatch/desync errors clearly.
5. **Export assertion ordering** — the static-shape assertion now runs **after** nncf compression (just before `save_model`), so int4/int8 NPU exports are covered too, not just the post-trace graph.
6. **fp32 + npu** — `export_shards.py` fails fast (`--default-dtype must be fp16 for --target npu`).
7. **present factorization** — `static_infer` validates each `present.*` output's **shape** `[1,kv_heads,context,head_dim]`, not just byte length.
8. **Empty prompt** — rejected with a clear error instead of emitting a fabricated token.
9. **Error-path ring reset** — `step_first`'s handler now calls `sk.reset()` alongside `reset_state()`.
10. **Per-token allocations** — cached resolved port names on `StaticKv`, a reused `mask_bytes` buffer (rewritten in place), position bytes on the stack, and dropped the per-token `outs` Vec. The decode hot loop no longer allocates per token.
11. **Altitude** — centralized the static-vs-stateful **wire protocol** into the `encode/decode_wire_position` pair (the single source of truth that makes #2/#3/#4 enforceable in one place). I deliberately did **not** do the full `KvStrategy` trait-object rewrite: it's a large refactor of freshly-validated multi-stage code, and the concrete risks it would mitigate (#2/#3/#4) are now directly fixed. Flagging it as a clean follow-up if you want it.
12. **Duplication** — extracted `resolve_next_token` (shared by the static loop + stateful arm); the wire framing is now one encode/decode pair.
13. **Redundant state** — `elem_bytes` is derived from `kv_dtype` (can't drift); the dead `static_seq` slot dropped from `static_params`. (`context` kept — it's set once at build as `past_len+1`; removing it was churn without the drift risk.)
14. **Dead code** — removed unused `f32_to_bytes`.

`cargo fmt` + `clippy` clean (only pre-existing dead-code lints remain). The single-stage CPU/NPU configs were re-validated as regressions to confirm the hardening didn't disturb the shipping path.

t8:
**Rebased onto `main` after #57 merged.** Resolved the `export_shards.py` overlap: the cpu-gpu path now uses #57's init-bearing `make_stateful_with_init` (CPU-loadable stateful KV), the `--target npu` path skips it (stateless static). Re-validated both export paths **12/12 on CPU** post-merge (qwen2.5-1.5B): cpu-gpu (init-bearing stateful, 28 KV pairs + cache_reorder) and npu-static (shape-check OK). runtime.rs unchanged by the rebase. Branch is a clean descendant of `main`.

Issue #60, opened 2026-06-02
cascadia.to advertises MoE families (Mixtral 8x7B, DeepSeek-MoE, Kimi K2) that shard can't export; main silently falls back to Llama and crashes


t8:
## Summary

cascadia.to's "Ships with support for…" list includes three families that the OSS `cascadia shard` pipeline **cannot** export:

- **"Mistral 7B · 8×7B"** — 8×7B is **Mixtral**, a sparse MoE.
- **"DeepSeek 7B–67B"** — the larger/realistic DeepSeek models (V2/V3) are MoE.
- **"Kimi K2 · 1T"** — MoE, and 1T params can't run on a single AI PC.

`docs/SHARDING.md` itself says "Mixtral / MoE — not supported," so the **website over-claims** relative to the code.

Worse, on `main` an unknown MoE arch isn't rejected — it's **silently treated as Llama** and then crashes:

```
warning: unknown model_type='mixtral', architectures=['MixtralForCausalLM'];
         falling back to Llama decoder layer (may break for non-Llama-compat models)
...
  File ".../export_shards.py", line 523, in build_wrapper
    layer = DecoderLayer(config, layer_idx=i)        # LlamaDecoderLayer
  File ".../transformers/models/llama/modeling_llama.py", line 234, in __init__
    self.scaling = self.head_dim**-0.5
TypeError: unsupported operand type(s) for ** or pow(): 'NoneType' and 'float'
```

## Reproduction

```
cascadia shard --model hf-internal-testing/tiny-random-MixtralForCausalLM \
  --output-dir ~/shards/mixtral --num-stages 1 --quantization int4 --python <venv>
```
(Used a tiny-random Mixtral to exercise the `MixtralForCausalLM` arch-dispatch path cheaply; the real Mixtral-8x7B is ~90 GB. The downstream `head_dim` crash is config-specific to the tiny model — the **reportable behavior is the silent Llama fallback for a MoE arch**, not the exact crash.)

## Already fixed on a branch (please merge)

`fix/exporter-arch-robustness` (`076d372`) implements exactly the right behavior: an explicit reject list with a **Mixtral-specific message** (`"Mixtral (model_type 'mixtral') is MoE …"`) and — importantly — it **detects + rejects before pulling the multi-GB model** (comments at lines 1331/1486). Merging it turns this crash into a clean, fast rejection.

## Asks

[...]

Issue #14, opened 2026-05-18
Surface sampling parameters in /v1/chat/completions (top_p, top_k, seed, stop, finish_reason, logprobs)


t8:
## Summary

The `/v1/chat/completions` endpoint accepts only `model`, `messages`, `max_tokens`, `temperature`, `stream` (`crates/tahoma-api/src/lib.rs:197-206`). Industry-standard fields are missing: `top_p`, `top_k`, `seed`, `stop`, `frequency_penalty`, `presence_penalty`, `logprobs`, `top_logprobs`. Streaming responses lack `usage` deltas, and `finish_reason` is always `"stop"` regardless of actual stop reason (`tahoma-api/src/lib.rs:417`).

`sparse-moe` already has `SamplingConfig` internally with `top_p` / `top_k` / `rep_penalty` / window. This issue surfaces those + several others at the HTTP API boundary and propagates them to all engines.

## Why

Tahoma's API hands non-default OpenAI clients an incomplete experience. Many evals (lm-eval-harness, llm-foundry, simple-evals) and downstream tools (LangChain, LlamaIndex, dspy, OpenAI Agents SDK) set these fields and expect them honored. Today they're silently dropped — clients have no way to tell that `top_p=0.9` they sent had zero effect.

`finish_reason` differentiation matters for observability — distinguishing "model finished naturally" (`stop`), "hit max_tokens" (`length`), "EOS token emitted" (also `stop` per spec), "explicit stop string matched" (`stop`), "client cancelled" (custom or `null`) is signal for routing decisions, retries, and quality checks.

### (1) Extend `ChatCompletionRequest`

**File:** `crates/tahoma-api/src/lib.rs:197-206`

Add fields:
```rust
pub struct ChatCompletionRequest {
    pub model: String,
    pub messages: Vec<ChatMessage>,
    #[serde(default)]
    pub max_tokens: Option<u32>,
    #[serde(default)]
    pub temperature: Option<f32>,
    #[serde(default)]
    pub top_p: Option<f32>,
    #[serde(default)]
    pub top_k: Option<u32>,
    #[serde(default)]
    pub seed: Option<u64>,
    #[serde(default)]
    pub stop: Option<StopSpec>, // string or array of strings, OAI-style
    #[serde(default)]
    pub frequency_penalty: Option<f32>,
    #[serde(default)]
    pub presence_penalty: Option<f32>,
    #[serde(default)]
    pub logprobs: Option<bool>,
    #[serde(default)]
[...]

PR #64, opened 2026-06-02
feat(shard): broaden exporter architecture support — Phi-3, Gemma-2, .bin loading, MoE rejection (#58/#59/#60/#61)

### LINKED ISSUE #58: shard: Phi-3 export fails — exporter uses split q/k/v but Phi3Attention has fused qkv_proj
## Summary

`cascadia shard` cannot export **Phi-3** models. During `torch.jit.trace`, the exporter's custom attention forward (`cached_layer_forward_sdpa`) accesses `layer.self_attn.q_proj`, but `Phi3Attention` exposes a **fused `qkv_proj`** (no separate `q_proj`/`k_proj`/`v_proj`):

```
File ".../export_shards.py", line 298, in cached_layer_forward_sdpa
    q = layer.self_attn.q_proj(hidden_states)
AttributeError: 'Phi3Attention' object has no attribute 'q_proj'. Did you mean: 'o_proj'?
```

`docs/SHARDING.md` lists Phi-3 (Mini, Medium) as "best-effort," and cascadia.to advertises "Phi 3.8B–14B," but Phi-3 never traces.

## Not a transformers-version artifact

Phi-3 has used a fused `qkv_proj` since it landed in `transformers`. Reproduced **identically** on:
- transformers **5.9.0** (latest on PyPI)
- transformers **4.49.0**

So this is an exporter gap, not a version regression. (Verified `pip index versions transformers`; 5.9.0 is latest.)

## Reproduction

```
cascadia shard --model microsoft/Phi-3-mini-4k-instruct \
  --output-dir ~/shards/phi3 --num-stages 1 --quantization int4 \
  --python <venv-with-torch+transformers+openvino+nncf>
```
Export gets through arch detection (`arch=phi`, 32 layers) and `Building wrapper...`, then fails in `torch.jit.trace` with the `AttributeError` above.

## Environment

cascadia `main` @ `fe392c9`, built `--features openvino`; OpenVINO 2026.1.0; export venv torch 2.12.0+cpu, nncf 3.2.0; transformers 4.49.0 **and** 5.9.0 both fail. No branch (incl. `fix/exporter-arch-robustness`) handles `qkv_proj`.

## Suggested fix

In `cached_layer_forward_sdpa`, detect a fused projection: if `self_attn` has `qkv_proj` (Phi-3, and similar archs) instead of `q_proj`/`k_proj`/`v_proj`, run the fused projection and slice it into q/k/v using `num_attention_heads`/`num_key_value_heads`/`head_dim` from config. Otherwise, add Phi to the explicit reject list so it fails fast with a clear message instead of an `AttributeError` mid-trace.

### LINKED ISSUE #59: shard: models without .safetensors load 0 tensors then KeyError (exporter is safetensors-only)
## Summary

The exporter's weight loader reads **`.safetensors` only** (`safe_open` over `*.safetensors`). For a model published with PyTorch `.bin` weights and **no** safetensors, it loads **zero tensors** and then dies with a misleading `KeyError` instead of a clear "no weights found" message:

```
Model: 30 layers, hidden=4096, kv_heads=32, rope_theta=10000.0, arch=llama
  Loading weights...
  0 tensors (0 MB)        <-- nothing loaded
  Building wrapper...
  ...
    embed_w = state_dict["model.embed_tokens.weight"]
KeyError: 'model.embed_tokens.weight'
```

The architecture detection succeeded (`arch=llama`); the problem is purely the weight **format**.

## Reproduction

```
cascadia shard --model deepseek-ai/deepseek-llm-7b-chat \
  --output-dir ~/shards/deepseek-7b --num-stages 1 --quantization int4 --python <venv>
```

The downloaded snapshot contains `pytorch_model.bin.index.json` (+ `.bin` shards) and **no** `.safetensors`:
```
config.json  generation_config.json  pytorch_model.bin.index.json  tokenizer.json  tokenizer_config.json
```

## Impact

cascadia.to advertises **DeepSeek 7B–67B**. The dense DeepSeek-LLM models (`deepseek-llm-7b-chat`, `-67b-chat`, `-base`) are distributed as PyTorch `.bin`, so they **fail out of the box** — and the error (`KeyError: 'model.embed_tokens.weight'`) looks like a model-structure bug rather than "I only read safetensors." Any `.bin`-only model is affected.

## Environment

cascadia `main` @ `fe392c9`; OpenVINO 2026.1.0; transformers 5.9.0. Confirmed the loader is `safe_open`-only on **all** current branches (`main`, `fix/exporter-arch-robustness`, `feat/standalone-sharding`, `docs/architectures-audit`, `feat/gemma4-cached-shards-port`).

## Suggested fix

Either (a) fall back to `torch.load` / `transformers` weight loading when no `.safetensors` are present (then proceed), or (b) at minimum detect "no safetensors found" up front and fail fast with an actionable message (e.g., "model ships pytorch_model.bin; re-save as safetensors or use a safetensors mirror") rather than a downstream `KeyError`.

### LINKED ISSUE #61: shard: Gemma-2 exports cleanly but produces CORRUPT output (missing embedding normalizer + Gemma-2 softcapping)
## Summary

**Gemma-2 exports without error but produces completely corrupt output.** `cascadia shard` on `unsloth/gemma-2-2b-it` (fp16) succeeds (exit 0, well-formed shard tree), the shard **loads and runs on the iGPU**, and then emits degenerate garbage for every prompt:

```
cap_fr →  Rov Rov Rov Rov Rov ... esprits esprits esprits ...
add    →  solaire solaire solaire ... følgeligfølgeligfølgelig ...
water  →  følgeligfølgeligfølgelig ... (x95)
```

Eval verdict: **0/12, CORRUPT** (every response is degenerate token repetition). This is the worst kind of failure — a server that looks healthy (`/health` OK, tokens flow at ~16 tok/s) but is numerically broken.

**It is Gemma-specific, not an fp16 or harness artifact:** Llama-3.2-1B exported via the *same* fp16 path and run through the *same* harness on the *same* GPU returns correct answers (11/12). Llama-3.1-8B, Qwen2.5-1.5B, Qwen3-1.7B, Mistral-7B (int4) all pass 11–12/12. Only Gemma is corrupt.

## Root cause

The generic exporter wrapper is Llama-shaped and omits Gemma's required ops:

- **Missing embedding normalizer.** `tools/export_shards.py` does `hidden_states = self.embed_tokens(input_ids)` (≈ lines 408 and 458) with no scaling. Gemma/Gemma-2 require `hidden_states *= sqrt(hidden_size)` (≈ 45.25× for hidden=2048). Without it the entire forward is off by ~45×.
- **Bypassed Gemma-2 softcapping + query scaling.** The hand-rolled per-layer forward `cached_layer_forward_sdpa` (≈ line 296+) uses a generic `scale = 1/sqrt(head_dim)` SDPA and the layer submodules directly, so Gemma-2's `attn_logit_softcapping`, `final_logit_softcapping`, and `query_pre_attn_scalar` never run. (It does import `Gemma2DecoderLayer`/`Gemma2RMSNorm`, but the custom forward doesn't call the layer's own `forward`.)

## Reproduction

```
cascadia shard --model unsloth/gemma-2-2b-it --output-dir ~/shards/gemma2-2b \
  --num-stages 1 --quantization fp16 --python <venv>      # succeeds
# ship shard to an Intel GPU box, then:
cascadia worker --rank 0 --total 1 --engine ov-runtime --device GPU \
  --model ~/shards/gemma2-2b --api 127.0.0.1:8000          # loads fine
# any prompt -> degenerate repetition garbage
```

## Docs / site mismatch

`docs/SHARDING.md` calls Gemma 2 "best-effort" with "logit softcapping NOT applied (~1% quality regression)" — but the real result is **total corruption, not ~1%**. cascadia.to advertises "**Gemma 2B–27B**" as supported.

## Already half-addressed (please finish)

`fix/exporter-arch-robustness` (unmerged) **detects** softcap and **rejects** Gemma-2 with a message ("the generic exporter cannot …"), and its comments explicitly list "`sqrt(hidden)` embed scaling" + "`attn/final_logit_softcapping` (Gemma 2)" as unsupported. So:

1. **Short term:** merge that branch so Gemma-2 fails *loudly* at export instead of silently producing garbage.
2. **To actually support Gemma (it's advertised):** apply the embedding normalizer and Gemma-2 softcapping / `query_pre_attn_scalar` in the exporter, then re-validate output.
3. Fix `docs/SHARDING.md` and cascadia.to to match whichever path is chosen.

## Environment

cascadia `main` @ `fe392c9`; OpenVINO 2026.1.0 (export) / GPU run on Arc 140V (Lunar Lake); transformers 5.9.0. Verified by greedy-decode eval with a degeneracy/corruption detector (caught here; missed by naive substring checks only if they ignore repetition).

t8:
## Broaden exporter architecture support

Re-ran the full model-family audit across CPU/GPU/NPU now that the device blockers (#57 CPU, #37 NPU) are fixed, and fixed the per-family export failures that remained. Each fix was researched from the model's HF architecture, not guessed.

Closes #58, closes #59, closes #61. Addresses #60 (clean rejection; full MoE export remains unimplemented).

### Fixes

**#58 — Phi-3 (fused qkv).** `cached_layer_forward_sdpa` only handled separate `q_proj/k_proj/v_proj`; Phi-3's `Phi3Attention` has a fused `qkv_proj`. Detect it and split by head sizes (`num_heads`/`num_kv_heads` × `head_dim`).

**#61 — Gemma-2 (silent corruption → correct).** Gemma-2 exported clean but emitted pure garbage (`Rov Rov Rov…`). The generic Llama-style forward was missing four Gemma-2 specifics, now carried by a new `ArchSpec`:
- token embeddings scaled by `sqrt(hidden_size)`;
- attention-logit tanh softcapping (`attn_logit_softcapping`) — SDPA can't do this, so attention is computed by hand when a softcap is set;
- final-logit tanh softcapping (`final_logit_softcapping`);
- query pre-attn scalar (`1/sqrt(query_pre_attn_scalar)` instead of `1/sqrt(head_dim)`);
- the 4-norm decoder structure (`post_attention_layernorm` *after* attention + `pre/post_feedforward_layernorm` around the MLP).

Gemma-1 gets the embedding scale only. (`head_dim` already came from config, so Gemma-2's 256≠2304/8 was handled.)

**#59 — DeepSeek-llm / legacy `.bin`.** The loader was safetensors-only; `deepseek-llm-7b-chat` ships `pytorch_model*.bin`. Added a `torch.load` fallback and `*.bin` to the download patterns. (deepseek-llm is Llama-arch, so once loaded it exports normally.)

**#60 — MoE rejection.** Mixtral / DeepSeek-V2-V3 / Qwen\*-MoE / Llama-4 silently fell back to a dense Llama layer and emitted garbage. `is_moe_config` now detects expert-count fields + known MoE model types and aborts up front with a clear message. Full MoE export (routing + per-expert MLPs) is not implemented — this just prevents a broken shard.

### Validation (CPU plugin, qwen2.5 eval suite, greedy, fp16/int4)

| Model | Before | After |
|---|---|---|
| Gemma-2-2B (#61) | 0/12 **CORRUPT** | **12/12**, corrupt=False |
| Phi-3-mini (#58) | export crash (AttributeError) | **11/12**, corrupt=False |
| deepseek-llm-7B (#59) | export crash ("0 tensors") | **11/12**, corrupt=False |
| tiny Mixtral (#60) | dense-fallback garbage | **clean rejection**, no shard |

The Phi/DeepSeek 11/12 misses are verbose answers truncated at `max_tokens=32` (e.g. the arithmetic-sequence prompt), not corruption. The fixes live in the shared export path, so they apply to the cpu-gpu (stateful) and — once #63 lands — the `--target npu` (static) export modes alike.

### Environment
cascadia `main` @ `ac9616f` (post-#57); OpenVINO 2026.1.0; transformers 4.57 / 5.9.

### DIFF
--- a/tools/export_shards.py
+++ b/tools/export_shards.py
@@ -27,18 +27,24 @@ for users who want fine-grained control:
 
 Architecture support: The script uses HuggingFace's AutoModelForCausalLM
 to load layers, so any model whose decoder layers expose the standard
-attribute names (`self_attn.{q_proj,k_proj,v_proj,o_proj}`,
-`mlp`, `input_layernorm`, `post_attention_layernorm`) and uses RoPE
-rotary will work. This covers:
-  - Llama 1/2/3/3.1/3.2 family
-  - Mistral 7B / Mixtral (single-expert per shard only)
-  - Qwen2 / Qwen2.5
-  - Some Phi-3 variants (those with standard attention)
-  - DeepSeek-V2-Lite
+attribute names (`self_attn.{q_proj,k_proj,v_proj,o_proj}` or a fused
+`qkv_proj`, `mlp`, `input_layernorm`, `post_attention_layernorm`) and uses
+RoPE rotary will work. This covers:
+  - Llama 1/2/3/3.1/3.2 family (+ Llama-arch models like deepseek-llm)
+  - Mistral 7B
+  - Qwen2 / Qwen2.5 / Qwen3
+  - Phi-3 (fused qkv_proj — #58)
+  - Gemma 1 (scaled embeddings) and Gemma-2 (+ attention/final-logit
+    softcapping + the 4-norm decoder structure — #61)
 
-Models with non-standard architectures (sliding-window attention with
-custom kernels, MoE routing per layer, multimodal) may export but won't
-match HF reference outputs; benchmark before trusting the result.
+Weights load from *.safetensors or, as a fallback, legacy pytorch_model*.bin
+(e.g. deepseek-llm — #59).
+
+Mixture-of-experts models (Mixtral, DeepSeek-V2/V3, Qwen*-MoE, Llama-4, …)
+are detected and rejected up front (#60): the exporter builds dense decoder
+layers and does not implement MoE routing / per-expert MLPs. Other
+non-standard architectures (custom sliding-window kernels, multimodal) may
+export but won't match HF reference outputs; benchmark before trusting.
 """
 
 import argparse
@@ -88,6 +94,25 @@ def detect_architecture(config) -> str:
         flush=True,
     )
     return "llama"
+
+
+def is_moe_config(config) -> bool:
+    """True if `config` describes a mixture-of-experts model.
+
+    cascadia's exporter builds dense decoder layers (one MLP per layer). MoE
+    models route each token through a subset of many expert MLPs per layer,
+    which this exporter does not implement. Detecting them lets the caller
+    reject early with a clear message instead of silently falling back to a
+    dense Llama layer and emitting garbage (#60).
+    """
+    for field in ("num_local_experts", "n_routed_experts", "num_experts"):
+        v = getattr(config, field, None)
+        if isinstance(v, int) and v > 1:
+            return True
+    model_type = getattr(config, "model_type", "").lower()
+    arch_first = ((getattr(config, "architectures", []) or [""])[0]).lower()
+    markers = ("mixtral", "moe", "deepseek_v2", "deepseek_v3", "llama4", "grok")
+    return any(m in model_type or m in arch_first for m in markers)
 
 
 def get_decoder_layer_cls(arch_tag: str):
@@ -282,12 +307,23 @@ def cached_layer_forward_sdpa(
     num_heads,
     num_kv_heads,
     head_dim,
+    attn_softcap=None,
+    query_scale=None,
+    gemma2_norms=False,
 ):
     """One decoder layer forward, attention via SDPA, KV-cached.
 
     Works for any decoder layer that exposes the conventional projection
-    names (`self_attn.q_proj/k_proj/v_proj/o_proj`, `input_layernorm`,
-    `post_attention_layernorm`, `mlp`).
+    names (`self_attn.q_proj/k_proj/v_proj/o_proj` or a fused `qkv_proj`,
+    `input_layernorm`, `post_attention_layernorm`, `mlp`).
+
+    Gemma-2 needs three deviations (#61), gated by the optional args:
+      - `query_scale`: attention scale = 1/sqrt(query_pre_attn_scalar) instead
+        of 1/sqrt(head_dim);
+      - `attn_softcap`: tanh-softcap the attention logits (SDPA can't, so we do
+        attention manually when set);
+      - `gemma2_norms`: the 4-norm residual structure (post_attention_layernorm
+        AFTER attn + pre/post_feedforward_layernorm around the MLP).
     """
     bsz, seq_len, _ = hidden_states.shape
     num_kv_groups = num_heads // num_kv_heads
@@ -295,9 +331,19 @@ def cached_layer_forward_sdpa(
     residual = hidden_states
     hidden_states = layer.input_layernorm(hidden_states)
 
-    q = layer.self_attn.q_proj(hidden_states)
-    k = layer.self_attn.k_proj(hidden_states)
-    v = layer.self_attn.v_proj(hidden_states)
+    if hasattr(layer.self_attn, "qkv_proj"):
+        # Fused QKV projection (e.g. Phi-3, #58): one matmul, split by size into
+        # q (num_heads), k/v (num_kv_heads), each * head_dim.
+        qkv = layer.self_attn.qkv_proj(hidden_states)
+        q_sz = num_heads * head_dim
+        kv_sz = num_kv_heads * head_dim
+        q = qkv[..., :q_sz]
+        k = qkv[..., q_sz : q_sz + kv_sz]
+        v = qkv[..., q_sz + kv_sz : q_sz + 2 * kv_sz]
+    else:
+        q = layer.self_attn.q_proj(hidden_states)
+        k = layer.self_attn.k_proj(hidden_states)
+        v = layer.self_attn.v_proj(hidden_states)
 
     q = q.view(bsz, seq_len, num_heads, head_dim).transpose(1, 2)
     k = k.view(bsz, seq_len, num_kv_heads, head_dim).transpose(1, 2)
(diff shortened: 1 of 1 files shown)


t8:
## Device validation — full family × device matrix (CPU / GPU / NPU)

Re-ran the audit across all three Intel device classes now that **#57 fixed CPU** (merged as #62) and **#37 is fixed by #63** (NPU static-shapes + multi-stage). Every cell below is a **real measured verdict** from the `cascadia worker` binary — greedy decode, 12-prompt eval suite, `int4`, single-stage unless noted. No cell is inferred from another.

- **CPU** = miner (Cascade Lake), `--device CPU`, post-#62 binary.
- **GPU** = pawan-01 Arc 140V iGPU, `--device GPU`.
- **NPU** = pawan-01 AI Boost NPU, `--device NPU`, static graph (`--target npu --static-context 256`), needs the #63 branch.

### Family × device matrix (pass-rate, all `corrupt=False`)

| Family | CPU | GPU (Arc 140V) | NPU (AI Boost) |
|---|---|---|---|
| Llama-3.2-1B | 10/12¹ | 11/12¹ | 11/12¹ |
| Qwen2.5-1.5B | 12/12 | 12/12 | 12/12 (+ 2-stage 12/12) |
| Qwen3-1.7B | 12/12² | 12/12² | 12/12² |
| **Gemma-2-2B (#61)** | **12/12** | **12/12** | **12/12** |
| Mistral-7B-v0.3 | 12/12 | 12/12 | 12/12 |
| **Phi-3-mini (#58)** | **11/12**¹ | **12/12** | **12/12** |
| **deepseek-llm-7B (#59)** | **11/12**¹ | **11/12**¹ | **11/12**¹ |
| Mixtral / MoE (#60) | clean reject | clean reject | clean reject |

¹ Sub-12 scores are **not corruption** — every output was correct and coherent. The misses are: Llama-3.2-1B (1B, weakest model) flubbing the arithmetic-sequence continuation and one trivia name; Phi-3/DeepSeek giving verbose answers that get cut at `max_tokens=32` on the sequence prompt. The grader counts a truncated-but-correct answer as a miss; none are degenerate/garbled.

² Qwen3-1.7B is a **hybrid-reasoning** model — it emits a `<think>…</think>` block before answering, so it needs a larger token budget to reach the final answer. All three Qwen3 cells use `max-tokens=160` (at 32 it scores 4/12 purely from think-block truncation — `corrupt=False`, every trace coherent and on-topic). The other (non-reasoning) families use `max-tokens=32`.

### What this resolves
- **Gemma-2 (#61) was the headline:** 0/12 **CORRUPT** (`Rov Rov Rov…`) on main → **12/12 on all three devices** after the embed-scale + attn/final softcapping + 4-norm fix.
- **Phi-3 (#58)** fused-qkv split and **deepseek-llm-7B (#59)** `.bin` loader: both export and run correctly across the devices tested.
- **MoE (#60):** config-first hard reject (no multi-GB download) on every device — Mixtral/DeepSeek-MoE/Kimi fail fast with an actionable message instead of silently building a broken dense shard.

### NPU has no practical size ceiling up to 7B (correcting an earlier note)
An earlier interim note in this work speculated a ~2.6B NPU compile ceiling (Phi-3-mini appeared to hang). **That was wrong** — it was process contention: a parallel 7B job's `Stop-Process` was killing the Phi-3 NPU worker mid-compile. On clean, **sequential** runs the NPU static graph compiles and runs correctly at **3.8B (Phi-3-mini, 12/12)** and **7.2B (Mistral-7B, 12/12)**. The only real effect of model size on NPU is **compile latency** (seconds for ~1.7B, a few minutes for 3.8–7B) — an operational characteristic, not a failure mode. No new issue warranted.

### Eval-harness footnote (not cascadia)
`eval_suite.py` initially crashed on Windows printing Gemma-2's `H₂O` — pawan's cp1252 console can't encode the `₂` subscript (`UnicodeEncodeError`). The model output was correct; re-running with `PYTHONIOENCODING=utf-8` gave the clean verdicts above. Linux/miner was unaffected.

t8:
## Code review — #64 (max-effort)

The **validated path is correct** (finders confirmed against the HF reference): the Gemma-2 manual attention (scale→softcap→mask→fp32-softmax), `query_scale = 1/sqrt(query_pre_attn_scalar)`, the 4-norm structure, embed scaling, final-logit softcap, and the Phi-3 fused-qkv split (incl. GQA sizing) all match. The findings below are about **claims that overreach the validated models**, and **robustness/efficiency** gaps.

### Correctness — claims broader than what works
1. **Gemma-1 export is broken despite the new claim.** `get_decoder_layer_cls('gemma')` / `get_norm_cls('gemma')` pick `Gemma2DecoderLayer`/`Gemma2RMSNorm` by import-order (only an `ImportError` fallback), not by `model_type`. On modern transformers the gemma2 module always imports, so a Gemma-1 model is built with `Gemma2DecoderLayer`, whose `Gemma2Attention.__init__` reads `config.query_pre_attn_scalar` — a field `GemmaConfig` lacks → **AttributeError**. The docstring now advertises "Gemma 1 (scaled embeddings)" but it can't export. *(Pre-existing class-selection bug, surfaced by the new claim.)* Fix: select layer/norm class by `model_type` (gemma vs gemma2), or drop the gemma-1 claim.
2. **partial-rotary / longrope ignored.** `apply_rotary` rotates the full `head_dim` and `TracedRotaryEmbedding` ignores `rope_scaling` (longrope). Phi-3-mini-4k (validated) is full-rotary so it's fine, but **phi-2/phi-1.5** (`partial_rotary_factor=0.4`) and **Phi-3.5-mini/Phi-4** (longrope) silently export garbage under the broad "Phi-3" claim. Fix: detect `partial_rotary_factor<1` / longrope `rope_scaling` and reject-or-warn (or implement), and narrow the doc claim to what's validated.
3. **Gemma-3 mis-handled (same root as #1).** `model_type == 'gemma2'` exact-match → gemma-3 falls into the gemma-1 branch (embed-scale only) and is built as a Gemma2 layer. Not claimed, but the same `model_type`-keyed selection fixes both.

### MoE rejection (#60) — gaps
4. **Downloads the whole model before rejecting.** `is_moe_config` runs *after* `maybe_download`, so Mixtral / DeepSeek-V3 / DBRX pull tens-to-hundreds of GB before the millisecond config-only reject — defeating the fail-fast intent. Read `config.json` (or `AutoConfig` on the repo id) before snapshotting weights.
5. **False negatives.** DBRX nests its expert count under `config.ffn_config.moe_num_experts` (top-level fields miss it) → falls back to a dense Llama layer and emits garbage — exactly what #60 set out to prevent. Also `num_experts == 1` (the `> 1` guard) and configs exposing only `num_experts_per_tok` slip through.
6. **False positives.** Bare substrings `'moe'`/`'grok'` match any `model_type`/arch *containing* them — a dense model named `*moe*` would be wrongly rejected. Anchor the markers.

### `.bin` loader (#59) — edges
7. **Bare `*.bin` fallback grabs non-weight files.** `glob("pytorch_model*.bin") or glob("*.bin")` — the fallback `torch.load`s `training_args.bin`/`optimizer.bin`/etc.; `weights_only=True` raises on those → export aborts. Restrict to `pytorch_model*`/`model*` patterns.
8. **`maybe_download` now pulls `*.bin` for every model.** Repos shipping both formats download a redundant (~16 GB for an 8B) `.bin` that `load_stage_weights` never reads (it prefers safetensors). Two-pass: fetch `*.bin` only when no safetensors arrived.
9. **Whole-shard load → OOM risk.** `torch.load` materializes the entire `.bin` shard in RAM before filtering to the stage's keys (vs the safetensors lazy `get_tensor`); a 7B `.bin` is ~6–7 GB resident per stage on the laptop-class targets cascadia aims at.
10. **`weights_only=True` has no fallback** — some legacy `.bin` need full unpickle; a `try/except` to `weights_only=False` (or a clear message) would keep the #59 fallback robust.

### Cheap guards / cleanup
11. **No shape assert on the qkv split** — `qkv.shape[-1] == q_sz + 2*kv_sz` is unchecked; a non-`[Q,K,V]` fused layout silently mis-splits. One `assert` makes it loud.
12. **Embed-scale + final-softcap are duplicated** across the two embed wrappers and two head wrappers (via replace_all) instead of `_BaseStage` helpers — drift risk (multi-stage Gemma could lose the embed scale on one stage and the single-box e2e test wouldn't catch it).
13. **Altitude:** arch knowledge is now split across `detect_architecture` + `get_decoder_layer_cls` + `get_norm_cls` + `arch_spec_from_config`, keyed three different ways; adding an arch is 4–5 coordinated edits. A per-arch registry (`tag → {layer_cls, norm_cls, spec}`) would make each arch one entry. `cached_layer_forward_sdpa` is also now a 5-branch multi-arch conditional.
14. **Stale top-of-file docstring** (line ~4) still says "Llama, Mistral, Qwen2, or any decoder-only LLM" — contradicts the new MoE hard-reject 40 lines below and omits Qwen3/Phi-3/Gemma.

**Net:** the three headline fixes (Phi-3-mini, Gemma-2-2B, deepseek-llm-7B) are correct and validated. The actionable items are #1 (gemma-1 false claim/crash), #4 (MoE download-before-reject), #5–#8 (MoE/`.bin` robustness), and #2 (narrow/guard the Phi claim) — most are small. #11–#14 are cleanup.

PR #65, opened 2026-06-03
feat: MiniMax-M2 support (single-stage OV-IR sparse-MoE)

t8:
## What

Adds **MiniMax-M2** support to Cascadia: a Python exporter that traces M2
into the sparse-MoE OV-IR layout, and an architecture-agnostic single-stage
runtime (`OvMoeRunner`) that runs it. Validated end-to-end against the
canonical HuggingFace model.

## Why M2 (and only M2), and why single-stage

I researched the whole MiniMax family first. Findings that shaped the scope:

- **No small MiniMax text model exists.** The smallest is M2 at **230B**
  total (~230GB FP8); Text-01/M1/VL-01 are 456B. Nothing fits the Intel
  AI-PC fleet (16–32GB/box) — even M2 at INT4 is ~115GB of weights before
  KV/activations, and "10B active" is a compute property, not memory (all
  256 experts/layer must be resident across the ring). So M2 runs
  **single-stage on a high-RAM node** with experts streaming from NVMe
  (the Kimi K2.6 approach), not pipeline-parallel across laptops.
- **M2 is the only tractable MiniMax text model.** Text-01/M1 use a
  lightning (linear) attention hybrid whose chunked recurrence doesn't
  trace into OpenVINO. M2 dropped lightning attention for plain
  full-softmax GQA, so it exports like a normal MoE transformer.

## Design: architecture-in-graph

The K2.6 sparse-MoE path runs a hardcoded MLA attention kernel in Rust —
incompatible with M2's GQA. Instead of porting a new AVX-512 kernel, this
PR restores the crate's **original OV-IR pipeline design** for M2: every
architecture detail (GQA, full-width QK-norm, partial RoPE, sigmoid routing
with `e_score_correction_bias` + renorm, SwiGLU experts) lives in the
exported OpenVINO graphs, so the Rust runtime stays dimension-agnostic and
correctness reduces to "does OV faithfully run what PyTorch traced."

`OvMoeRunner` (new `ov_moe.rs`): embed → per-layer shell (attention+routing,
KV threaded through a 7-tensor contract) → top-k expert dispatch + combine
→ head. Selected by `manifest.shell_backend == "ov_ir"`; the K2.6 MLA path
is untouched. `total > 1` is rejected (pipeline-parallel MoE isn't wired).

See `docs/MINIMAX_M2.md` for the full contract, export, and run steps.

## Validation — green

A `--tiny` synthetic M2 (faithful arch, random weights) is exported and run
through **both** the Python OV pipeline and the Rust `OvMoeRunner`, on a
real OpenVINO 2026.1 CPU node. Both reproduce the **canonical HF greedy
output exactly**:

```
reference greedy: [1, 2, 3, 4, 5, 231, 231, 229, 231, 229, 231]
ours     greedy : [1, 2, 3, 4, 5, 231, 231, 229, 231, 229, 231]
test result: ok. 1 passed
```

This exercises GQA, QK-norm, partial RoPE, sigmoid routing, expert SwiGLU,
and the token-by-token KV pipeline. The Rust test (`minimax_m2_eval`) is
gated on `M2_MODEL_DIR` and skips cleanly without OpenVINO/the fixture.
(Validation uses a `--no-quant` tiny export — INT4 on a 2-layer *random*
model is too lossy to preserve argmax; INT4 is exercised on the learned
full-model weights.)

## Not yet done (follow-up)

- **Full 230B run.** The FP8 checkpoint is downloading to the test node
  (~230GB, resumable, many hours); full INT4 export + a single-stage
  generation smoke test on the real weights is the remaining step. The
  exporter has the FP8→INT4 dequant path, but the fused-3D-expert FP8 scale
  handling only gets exercised once the real weights land.
- Greedy-only generation in v1 (sampling config ignored by `OvMoeRunner`).

## Test

```bash
python tools/export_minimax_m2.py --tiny --no-quant --out /tmp/m2_tiny
python tools/test_minimax_m2_pipeline.py --model-dir /tmp/m2_tiny
M2_MODEL_DIR=/tmp/m2_tiny cargo test -p cascadia-engine-sparse-moe \
  --features openvino --test minimax_m2_eval -- --nocapture
```


t8:
### Full 230B MiniMax-M2 — validated single-stage on the miner ✅

The remaining follow-up is done. Pipeline:

1. **Download** — `MiniMaxAI/MiniMax-M2` FP8, 215G / 130 shards, to the miner's external SSD.
2. **Streaming INT4 export** — `export_minimax_m2.py --model` (per-layer, RAM-bounded): 62 shells + **15,872 expert IRs**, 145G, ~3h.
3. **Single-stage run** via the actual `OvMoeRunner` (`minimax_m2_full_generate_smoke`):

```
prompt:     "The capital of France is"
completion: " Paris. The capital of the United Kingdom is London. capital capital both both thérapeut Hedge mainland United both"
20 tokens in 469.7s (~0.04 tok/s, CPU, unoptimized OV-IR expert path) — test PASSED
```

**The model produces the correct answer** — *"Paris"* and *"London"* with correct sentence structure. Since wrong rotary, routing, expert mapping, or KV handling would yield immediate garbage rather than correct multi-fact recall, this validates the export + engine on real learned weights end-to-end.

A real bug was caught and fixed *before* the multi-hour export: the M2 `config.json` carries a top-level `rotary_dim: 64` (no `rope_parameters`), so a raw-JSON read defaulted to full rotary; loading via `AutoConfig` gives the correct `partial_rotary_factor = 0.5 → rotary_dim 64`. FP8→INT4 dequant was also spot-checked (sane magnitudes, no NaN).

**Known limitation (follow-up):** the completion degrades into repetition after the correct first sentence — expected for greedy-only v1 (no repetition penalty) and likely compounded by INT4 expert noise. Next steps: repetition penalty / sampling in `OvMoeRunner`, and an INT8-expert comparison. Throughput (~0.04 tok/s) is unoptimized — the OV-IR per-expert dispatch compiles/runs each expert through the OV CPU plugin; the int4_bin AVX-512 expert path (as used for K2.6) is the obvious optimization.

Reproduce:
```bash
cascadia worker --engine sparse-moe --model <m2_int4_export> --device CPU
```

t8:
### Both caveats addressed + measured (full 230B M2, miner)

#### Performance — int4_bin AVX-512 expert backend

Added an `int4_bin` expert backend (flat int4-packed experts, group-size
32, fed to the `cascadia-int4-gemm` AVX-512 kernel) as an alternative to
per-expert OpenVINO-plugin calls. The kernel derives dims from slice
lengths, so it's reused as-is for M2 (3072/1536). Measured warm
steady-state (24 tokens, identical fp32 shells, cold first-step excluded):

| Expert backend | warm tok/s | first decode-step |
| --- | --- | --- |
| ov_ir (per-expert OV plugin) | ~0.063 | ~18–23 s |
| **int4_bin (AVX-512 kernel)** | **~0.21** | **~6–7 s** |

**≈ 3.2× faster warm, ≈ 3.5× faster first token, equal output quality.**
Convention validated by a Rust unit test (`int4_bin_expert_matches_fp32`,
rel-L2 0.08 vs fp32) and a Python packer round-trip (0.10).

#### Quality — sampling + routing precision

`OvMoeRunner` now generates via `crate::sampling` (repetition penalty /
temperature / top-p); greedy stays the default (tiny correctness test
still matches HF exactly). Comparison on the prompt "The capital of France
is":

| Config | completion |
| --- | --- |
| greedy, int4 router | ` Paris. … London. capital capital both both thérapeut …` |
| **rep-penalty 1.3** | ` Paris. … London.` then no literal loop, but still drifts |
| **rep-penalty + fp32 router/lm_head** | ` Paris. The capital of the United States is Washington, …` then drifts |

Findings:
- **Repetition penalty** removes the literal `capital capital both both`
  loops greedy falls into.
- **Precise routing** (`--shell-quant none --head-quant none` — M2 keeps
  the gate / `e_score_correction_bias` / lm_head full-precision; my export
  was INT4-ing them) extends coherent output from one fact to two.
- The residual degradation (incoherent, often multilingual, after ~12–14
  tokens) is dominated by the **int4 experts** — the bulk of the model at
  ~4 bits. The model recalls facts correctly but loses sustained
  coherence; restoring it needs int8 experts or calibrated quant
  (AWQ/GPTQ-style). int8 experts ≈ 230 GB, beyond the test box's disk —
  documented as future work.

Net: the model produces correct factual output; repetition is fixed,
routing precision improved, and the int4_bin path makes decode ~3× faster
at no quality cost. New exporter knobs: `--experts int4_bin|ov_ir`,
`--shell-quant`/`--head-quant none|int8|int4`. See `docs/MINIMAX_M2.md`.

t8:
### int8 experts — proven to fix coherence (full int8 artifact is disk-bound)

**The full int8 expert artifact doesn't fit the test box:** int8 experts are
~225 GB (8-bit × the 224.8B expert params) and the FP8 source (215 GB,
required *during* quantization) can't coexist with it on the 448 GB disk —
even after freeing the int4_bin experts (→ ~220 GB free, still short with no
headroom).

So I tested the underlying question more decisively, with **zero extra
disk**: run the experts **unquantized (fp32, dequantized from the FP8
source)** through the exact same exported OV shells/embed/head
(`tools/test_minimax_m2_fp_experts.py`). fp32 experts is the *ceiling* —
strictly better than int8.

| Expert precision | completion ("The capital of France is") |
| --- | --- |
| int4 (group-32) | ` Paris. … London. capital capital both both thérapeut Hedge 三省 …` |
| int4 + fp32 router/lm_head | ` Paris. The capital of the United States is Washington, … 三省 Canada …` |
| **fp32 (unquantized)** | **` Paris. The capital of the United States is Washington, D.C.\n\nThe capital city (or simply "capital") refers …`** |

**fp32 experts → fully coherent, no degradation.** This *confirms* the int4
experts are the coherence ceiling — not routing, attention, RoPE, or a bug
(those are identical across all three rows). Since int8 lies between int4
and fp32 (and far closer to fp32), **int8 experts would restore coherence**.

**Conclusion for the quality caveat:**
- repetition → fixed (repetition penalty),
- routing precision → fixed (fp32 gate/lm_head),
- sustained coherence → gated by expert precision; fp32 experts prove the
  fix. Shipping it as an int8 *artifact* needs a larger disk (the exporter's
  `--experts-int8` path produces int8 OV-IR experts where disk allows), or
  calibrated int4 (AWQ/GPTQ-style) to keep the int4 footprint while
  recovering quality. Documented in `docs/MINIMAX_M2.md`.

t8:
### Built the real int8 expert artifact — and it does *not* restore coherence

Freed the disk (my scratch only — left the K2.6 artifacts alone) and built
the full int8 artifact in place: **15,872 int8 OV-IR experts** (NNCF
int8_sym per-channel), via a new exporter mode `--free-source-shards` that
deletes each FP8 expert shard once its last consumer layer is exported
(experts are layer-contiguous, so peak disk stays flat). Then ran it.

**Surprising result — int8 degrades like int4, not like the fp32 probe**
(everything else identical: routing, attention, fp32 shells):

| experts | "The capital of France is" → |
| --- | --- |
| **fp32** (FP8-precision) | ` Paris. The capital of the United States is Washington, D.C. The capital city …` ✅ |
| **int8** (per-channel linear) | ` Paris. … Washington, capitals … 三省三省 . United . and Â 糊 base` ❌ |
| int4 (group-32 linear) | ` Paris. … Washington, capitals … 三省三省 United .. 三省 and 俞` ❌ |

So my earlier "int8 would restore coherence" prediction was **wrong**, and
that's the interesting part: it's **the *kind* of quantization, not the bit
width.** The experts are natively **FP8 (e4m3 — float, log-spaced, wide
dynamic range)**. The fp32 probe preserves that exactly → coherent. But
**linear** int8/int4 (uniform step max/127 or max/7) collapses the
small-magnitude weights FP8's log spacing kept → both degrade, int8 ≈ int4.

**Takeaway:** more *linear* bits doesn't help. The fix is a
**distribution-matched / float quant** — FP8 experts (keep the native
format), NF4, or AWQ/GPTQ-style calibrated int4. A plain int8 artifact is
now built and runs (`--experts-int8`) but isn't the quality answer.

Honest cost note: building int8 in place consumed the FP8 source (deleted
progressively, by design, to fit on disk). Re-dissecting this (e.g. the
int8-*simulate* probe, now wired as `--quant int8`) needs a fresh ~230 GB
FP8 re-download. Recommended next step for quality is an **FP8 or NF4
expert backend** rather than linear int8/int4. Docs: `docs/MINIMAX_M2.md`.

t8:
### Resolved: MiniMax-M2 generates correctly — fast and coherent ✅

The "degrades after the first sentence" was **not** expert quantization. Root
cause: the **OpenVINO 2026.1 CPU snippets shape-specialization bug**. The
shells run as OV graphs whose `past_k`/`past_v` dimension grows by one each
token; that per-step shape change triggers a snippets code-gen bug whose
numerical error accumulates and corrupts output after ~10 tokens. **One-line
fix:** `SNIPPETS_MODE=DISABLE` on the CPU plugin in `OvMoeRunner` (the
exporter/Python reference always set it; K2.6 dodges it by running shells in
its Rust kernel).

**Final result — full 230B M2, int4 experts + AVX-512 int4_bin + fix:**

```
greedy:      " Paris. The capital of the United States is Washington, D.C. The capital of the United Kingdom is London. The"
rep-penalty: " Paris. The capital of the United States is Washington, D.C. The capital city (or simply "capital") refers"
~0.19 tok/s warm (≈3–4× ov_ir), first decode-step ~5–7 s
```

Fully coherent through all 24 tokens, at the **smallest/fastest** config
(int4 group-32, ~115 GB). int4 was always fine.

**How it was found (and an honest retrace):** I initially misread this as a
quantization-precision problem and went int4 → int8 → NF4 → FP8-research,
each "confirming" the experts were the issue — because *every quantized run
was the Rust engine* (all degraded) while *every coherent run was the Python
reference*. The fix came from A/B-ing the **same OV graphs** through the two
pipelines: Python coherent, Rust degraded ⇒ engine config, not the weights.
Tooling added along the way (kept, they're useful): a `--quant
{none,int8,int4,nf4}` weight-quant simulator and a `--ov-experts` mode in
`tools/test_minimax_m2_fp_experts.py`, plus `--expert-quant`/`--shell-quant`/
`--head-quant` and `--free-source-shards` in the exporter. **Lesson (now in
the docs): when int8 degrades exactly like int4, suspect a systematic
pipeline bug, not precision — A/B the same artifact across pipelines before
re-quantizing anything.**

Net for the PR: `cascadia worker --engine sparse-moe --model <int4_bin export>
--device CPU` runs the full 230B MiniMax-M2 single-stage, coherently, at
~0.19 tok/s on the Xeon. Docs updated; tiny correctness test still green.

PR #69, opened 2026-06-03
feat(shard): partial rotary support + config-first arch rejection (reconcile #47)

t8:
Reconciles the still-unique value of #47 onto current `main`, and closes #47.

> **Update:** a thorough multi-angle review of the first push caught a **critical bug** in the ported partial-rotary code (it matched HF's cos/sin *values* but applied the rotation with the wrong dim pairing — `max|Δ|=7.3` on the applied q/k, ~88% of dims wrong, silently mis-exporting the models it claimed to support). It has been **fixed** (rewritten to HF's contiguous-slice convention) and re-validated against the **applied** q/k. See the review comment below.

## Why
#47 was opened on a pre-#64 base. Since then #64 merged the config-first MoE rejection (`is_moe_config`) and the Gemma-1/2 distinction. This PR pulls forward only the parts of #47 that #64 did **not** subsume, adapted to main, and **drops the now-redundant MoE-family rejections** (already covered by `is_moe_config`).

## What

**Partial rotary** (Phi-1/2, StableLM, Persimmon, Phi-4-mini) — replaces main's hard rejection.
- `TracedRotaryEmbedding` emits cos/sin at width `rotary_dim = int(prf·head_dim)` (not `head_dim`), and `apply_rotary` rotates **only the leading `rotary_dim` slice** of each head (`rotate_half` splitting at `rotary_dim/2`), concatenating the contiguous tail untouched — exactly transformers' Phi/Phi3 `apply_rotary_pos_emb`. **Verified against the APPLIED q/k byte-for-byte** on transformers 4.57 and 5.9. Odd `rotary_dim` and `partial_rotary_factor=0` (NoPE) are rejected, not coerced.
- Threaded through the stage wrappers / `build_wrapper` / `export_single_stage` / configs; `--partial-rotary-factor` override.

**Config-first rejection** (`detect_architecture`, before any download):
- **Gemma 4** — the headline bug: `model_type "gemma4"` *contains* `"gemma"`, so on main it falls through to the Gemma-1 path and is silently mis-exported. Now raises a typed `UnsupportedModelError`.
- **Gemma 3**, **gpt-oss**, **Mamba/hybrid** (pure Mamba isn't MoE, so `is_moe_config` misses it). Multimodal `text_config` is unwrapped first.
- `CASCADIA_ALLOW_LOSSY_EXPORT=1` escape hatch; raw-`config.json` fallback for model types transformers itself can't parse.

**`check_export_quirks`** severity gate — HARD (abort: asymmetric head_dim, per-layer embeddings, KV sharing, QK-norm off the qwen3 path, non-Gemma-2 softcap, plus a MoE catch that reuses `is_moe_config`) vs SOFT (warn + proceed: long-context RoPE scaling, sliding window). Gemma-2 softcap and Qwen3 QK-norm are **not** flagged (#64 handles them).

**Multimodal consistency** — `main()` collapses to `_text_config(config)` once before reading per-stage fields, so a wrapper whose backbone nests under `text_config` reads the right dims (detection already unwrapped). `_RawConfigNS` guards `_`-prefixed attribute access against copy/pickle recursion.

**Tests** — `tools/tests/test_arch_detection.py`, 63 cases: accept/reject matrix, `is_moe_config` as the MoE gate, HARD/SOFT classification incl. the Gemma-2-softcap-not-flagged guard, and partial rotary verified by **applying the rotation** and comparing q/k to HF `apply_rotary_pos_emb` (full + partial head_dim/prf cases), plus odd-dim and NoPE rejection.

## Validation (miner CPU; pawan export venv for the cross-version rotary check)
- **63/63** unit tests.
- Partial-rotary **applied q/k** byte-identical to HF (`max|Δ|=0.0000`); full rotary unchanged.
- **Phi-4-mini** (partial rotary, rejected by main today) exports + runs **coherently** on CPU (before the fix it collapsed into `"It is the capital of France. It is the capital..."`; after, three real sentences about Paris).
- Dense **Qwen2.5-1.5B** regression: exports + runs + coherent (full-rotary path unchanged).
- Rejection on **real** configs: `google/gemma-4-E2B-it` → Gemma-4 reject (exit 2); `Mixtral-8x7B` → MoE reject.

Closes #47.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,11 @@ target/
 *.pdb
 # Cargo.lock IS tracked (binary project — pinned for reproducibility)
 
+# Python (tools/ exporter + its tests)
+__pycache__/
+*.pyc
+.pytest_cache/
+
 # Cascadia scratch dirs (bench logs, ad-hoc experiments)
 tmp/
 experiments/
--- a/tools/export_shards.py
+++ b/tools/export_shards.py
@@ -37,15 +37,25 @@ RoPE rotary will work. This covers:
   - Phi-3 (fused qkv_proj — #58)
   - Gemma 1 (scaled embeddings) and Gemma-2 (+ attention/final-logit
     softcapping + the 4-norm decoder structure — #61)
+  - Partial-rotary models (Phi-1/2, StableLM-2, Persimmon): only the
+    leading `partial_rotary_factor * head_dim` dims are rotated.
 
 Weights load from *.safetensors or, as a fallback, legacy pytorch_model*.bin
 (e.g. deepseek-llm — #59).
 
-Mixture-of-experts models (Mixtral, DeepSeek-V2/V3, Qwen*-MoE, Llama-4, …)
-are detected and rejected up front (#60): the exporter builds dense decoder
-layers and does not implement MoE routing / per-expert MLPs. Other
-non-standard architectures (custom sliding-window kernels, multimodal) may
-export but won't match HF reference outputs; benchmark before trusting.
+Unsupported models are rejected up front (before the multi-GB download)
+with a specific message rather than silently emitting garbage:
+  - Mixture-of-experts (Mixtral, DeepSeek-V2/V3, Qwen*-MoE, Llama-4, gpt-oss,
+    …) — is_moe_config (#60): the exporter builds dense decoder layers.
+  - Gemma 3 / Gemma 4 — interleaved local/global attention, per-layer-type
+    RoPE, KV sharing, per-layer embeddings (detect_architecture). Note
+    "gemma4" contains "gemma", so it is rejected explicitly before the
+    Gemma-1 path catches it.
+Lossier-but-runnable quirks (long-context RoPE scaling, sliding-window
+attention) warn and proceed (correct within the original context window);
+features that corrupt even short prompts (asymmetric head_dim, per-layer
+embeddings, KV sharing, QK-norm off the qwen3 path) abort unless
+CASCADIA_ALLOW_LOSSY_EXPORT=1 (check_export_quirks).
 """
 
 import argparse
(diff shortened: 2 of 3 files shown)


t8:
## Self-review (max-effort, 9 finder angles) — findings + fixes

A thorough review of the first push (`6ef6c3a`) caught a critical bug and several smaller issues. All correctness findings are fixed in `fa64a5d`.

### 🔴 CRITICAL (fixed) — partial-rotary apply was wrong
The ported code zero-padded `inv_freq` to `head_dim` width and rotated the **full** head with one `rotate_half` split at `head_dim/2`. That made the cos/sin *values* match HF (so my first "byte-parity" check passed), but it paired each dim with the **wrong partner** for `prf<1`: HF rotates the contiguous `[0:rotary_dim]` slice (pairing `i ↔ i+rotary_dim/2`) and passes `[rotary_dim:]` through. Comparing the **applied** q/k to HF `apply_rotary_pos_emb`: `max|Δ|=7.3`, **112/128 dims wrong** — silently mis-exporting exactly the models the PR claims to support, *replacing* the old hard `sys.exit` guard with garbage.
**Fix:** cos/sin now width `rotary_dim=int(prf·head_dim)`; `apply_rotary` slices `q_rot=q[...,:rotary_dim]`/`q_pass`, rotates only the slice (`rotate_half` at `rotary_dim/2`), concatenates the tail — matching HF. Applied q/k now byte-identical (`Δ=0`); full rotary unchanged.

### 🟠 HIGH (fixed) — tests gave false confidence
The two partial-rotary tests compared only cos/sin tensors / asserted the buggy passthrough layout, never applying the rotation. **Fix:** tests now apply `apply_rotary` and compare q/k to HF across head_dim/prf cases (the decisive check), assert HF's contiguous tail passthrough, and reject odd-`rotary_dim` / NoPE.

### 🟡 MEDIUM (fixed)
- `_text_config` was unwrapped for detection but **not** for the per-stage field reads → a multimodal wrapper would read dims off the outer config. Now `main()` collapses to `_text_config(config)` once before the field reads.
- `partial_rotary_factor=0.0` (NoPE) was coerced to `1.0` by `... or 1.0`. Now preserved → rejected with a clear error.
- Odd `int(prf·head_dim)` desynced from HF (count + denominator). Now uses `rotary_dim=int(prf·head_dim)` and rejects odd.

### ⚪ LOW (fixed/hardened)
- `_RawConfigNS.__getattr__` recursion on copy/pickle → guarded `_`-prefixed names.
- ValueError fallback: removed the dead `ALLOW_LOSSY` branch (transformers can't build an unparseable config); clearer actionable error.
- `check_export_quirks` MoE re-scan now reuses `is_moe_config` (single source of truth, no drift).

### ✅ Verified clean
RoPE is baked into the v5 OV IR at export (runtime feeds only ids/mask/pos), so the exporter fix is sufficient — no Rust change needed; the new JSON field is harmlessly ignored. `detect_architecture`'s new `raise` is caught at all call sites; `partial_rotary_factor` threading lands in the correct slots; no efficiency findings.

### Known limitations (not fixed; out of scope)
- The unsupported-family gate is partly by-name (`gpt_oss`/`gemma4`/`gemma3`/`mamba`) — a genuinely new family name still falls through to the Llama path with a warning.
- A model with a **uniform** sliding window and no `layer_types` list (e.g. Mistral-7B-v0.1) trips no quirk (correct within the window; a blanket warning would be noisy on supported Mistral/Qwen).

t8:
## iGPU + NPU validation (pawan-01, Lunar Lake: Arc 140V iGPU + Intel AI Boost NPU)

Re-validated the partial-rotary fix on real Intel accelerators before merge. Exported Phi-4-mini (`partial_rotary_factor=0.75`) with the fixed exporter — both a `cpu-gpu` (dynamic stateful) shard for the iGPU and a `--target npu` (stateless static; "NPU static-shape check OK: 67 inputs + 65 outputs static") shard — and ran each through `cascadia run --engine ov-runtime` (greedy, temp=0).

| device | tok/s | output |
|---|---|---|
| **iGPU** (Arc 140V) | 26.3 | "Paris is the capital of France and is renowned for its rich history and culture. It is often referred by the people as the 'City of Light' and is noted for its numerous landmarks including the Eiffel Tower and the Louvre. Paris is noted" |
| **NPU** (AI Boost) | 6.3 | *(byte-identical to the iGPU output)* |

Both **coherent**, no repetition collapse — and the NPU output is byte-identical to the iGPU's (deterministic decode), confirming the same correct rotary is baked into the static NPU IR. Together with the CPU run (miner) and the 63 unit tests (applied-q/k byte-parity vs HF), partial rotary is now confirmed on **all three device classes (CPU, iGPU, NPU)**. Ready to merge.

PR #70, opened 2026-06-03
feat(cascadia-api): expose render_chat_prompt for in-process embedders

pawanpaudel93:
## What
Add `pub fn cascadia_api::render_chat_prompt(cfg, messages)` — a thin public wrapper over the existing (private) `render_prompt_with_template` / `render_prompt_legacy`.

## Why
In-process embedders that embed the engine directly bypass cascadia-api's HTTP handler, so they can't reach the HF chat-template renderer — only the `pub load_chat_template_config` loader was exposed. Such callers fall back to last-message-content, which makes Instruct models behave like base models (raw completions, dropped system/multi-turn context).

This exposes the same renderer the HTTP server uses (minijinja pycompat + `raise_exception` + `strftime_now` intact), so embedders get identical prompt formatting via `load_chat_template_config(model_dir)` → `render_chat_prompt(cfg, messages)`.

## Notes
- Purely additive; `render_prompt_with_template` / `render_prompt_legacy` stay private; no change to the HTTP server path.


7i7o (approved):
tACK 👍

PR #72, opened 2026-06-06
fix(gemma4-export): asymmetric global KV heads + k_eq_v for Gemma-4-31B

t8:
## Problem
`tools/export_gemma4.py` crashed exporting the current **Gemma-4-31B-it** (the doc's "31B tested" predated config changes). Two distinct bugs, both only hit by 31B (E2B/E4B don't have these features):

1. **Asymmetric global KV heads.** 31B's `full_attention` (global) layers use `num_global_key_value_heads` (4) vs `num_key_value_heads` (16) for sliding layers — alongside the asymmetric `head_dim` (256/512). The exporter threaded only per-layer `head_dim`, not per-layer KV-head count.
   `RuntimeError: shape '[1, 4, 16, 512]' is invalid for input of size 8192`
2. **`attention_k_eq_v`.** 31B global layers set `v_proj = None` and derive V from the raw `k_proj` output (HF `Gemma4TextAttention`: `value_states = key_states` *before* k_norm/rope), then `v_norm`. The exporter unconditionally called `v_proj`.
   `TypeError: 'NoneType' object is not callable`

## Fix
- Thread a per-layer `num_kv_heads` list through the wrapper forward + example KV inputs (mirrors the existing per-layer `head_dims`), so the stateful KV Variables get the correct global-layer shape.
- When `v_proj is None`, derive V from the pre-norm raw `k_proj` output (then `v_norm`), matching HF.
- E2B/E4B unaffected (no global-KV field / `k_eq_v=false` → fallbacks preserve prior behavior).
- **Guard a present-but-`None`/`0` global-KV field** (commit `91ff356`): `getattr(text_config, "num_global_key_value_heads", default)` only falls back when the field is *absent*; a config that carries it but leaves it `None`/`0` would otherwise pass that through as the global KV-head count. Append `or num_kv_heads` at both sites so a falsy value also falls back to the local count. No-op for 31B (field = 4); belt-and-suspenders for other Gemma-4 variants.

## Validation
Gemma-4-31B-it int4 3-stage **exports + self-verifies** (Prefill/Decode OK per stage) and **runs end-to-end** pipeline-parallel across 3 Intel Arc iGPUs (~1 tok/s, coherent output).

**Re-validated on the NUC fleet, 2 × Intel iGPU, 2026-06-24.** A fresh **2-stage** int4 31B export from this branch (both stages self-verify Prefill/Decode OK) deployed `stage_0` → one NUC, `stage_1` → another, run as `cascadia worker --engine gemma4 --device GPU` over the TCP relay (cross-machine). Coherent output across prompts — e.g. *"…Capital city.\nParis."* and *"…17 + 25 = 42."*

Engine-measured decode (rank-0 `gemma4 task done` lines): **~2.0–2.5 tok/s**, single request / batch 1. Per-token cost ≈ `alpha_ms` (stage_0 iGPU compute) + `wire_ms` (relay + stage_1 compute, dominated by the downstream stage rather than the ~21 KB/token hidden-state transfer); representative 32-token request `alpha_ms≈7800 wire_ms≈6500`. Concurrent requests **do not** raise throughput on this engine: it uses a **stateful single-sequence KV cache** and runs one task at a time, so requests serialize (measured 3-concurrent = ×0.93; engine log shows strictly back-to-back `task active`/`task done` intervals). True throughput scaling would need continuous batching / in-flight microbatches (not implemented, and non-trivial with stateful KV). Net: 2-stage PP here buys **fitting 31B across two iGPUs**, not single-request speed or concurrency throughput.

### DIFF
--- a/crates/cascadia-api/src/lib.rs
+++ b/crates/cascadia-api/src/lib.rs
@@ -533,6 +533,15 @@ impl ChatPromptRenderer {
     }
 }
 
+/// One-shot chat-prompt render for standalone / embedded callers that hold a
+/// [`ChatTemplateConfig`] directly (e.g. an embedded shard backend) and render
+/// outside the router's `AppState`. Builds a [`ChatPromptRenderer`] per call;
+/// callers that render repeatedly should construct one [`ChatPromptRenderer`]
+/// and reuse it. Falls back to [`render_prompt_legacy`] on parse/render failure.
+pub fn render_chat_prompt(cfg: &ChatTemplateConfig, messages: &[ChatMessage]) -> String {
+    ChatPromptRenderer::new(cfg).render(messages)
+}
+
 async fn chat_completions(
     State(state): State<AppState>,
     Json(req): Json<ChatCompletionRequest>,
--- a/tools/export_gemma4.py
+++ b/tools/export_gemma4.py
@@ -252,11 +252,17 @@ def cached_gemma4_layer_forward(
 
     k = layer.self_attn.k_proj(hidden_states)
     k = k.view(bsz, seq_len, num_kv_heads, head_dim)
+    # k_eq_v (Gemma 4 31B global/full_attention layers): v_proj is None and V
+    # derives from the RAW k_proj output — HF sets value_states = key_states
+    # BEFORE k_norm/rope are applied. Capture v from the pre-norm k here.
+    if layer.self_attn.v_proj is not None:
+        v = layer.self_attn.v_proj(hidden_states)
+        v = v.view(bsz, seq_len, num_kv_heads, head_dim)
+    else:
+        v = k
     k = layer.self_attn.k_norm(k)
     k = k.transpose(1, 2)
 
-    v = layer.self_attn.v_proj(hidden_states)
-    v = v.view(bsz, seq_len, num_kv_heads, head_dim)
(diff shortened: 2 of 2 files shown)


t8:
Rebased onto latest \`main\` (\`145694c\`) to resolve the conflict — force-pushed \`06b708c\` → \`1593d94\`.

**What changed in the rebase:**
- The `fix(gemma4-export)` commit cherry-picked cleanly (main never touched `tools/export_gemma4.py`) — byte-identical to before.
- The `feat(api): expose render_chat_prompt` commit conflicted: since this branch was cut, main landed the chat-template refactor that **removed** the old `render_chat_prompt`, added a 5th arg to `render_with_chat_env`, and introduced a public **`ChatPromptRenderer`** (the AppState-free renderer for in-process callers — exactly this commit's use case). Rather than re-introduce a near-duplicate that no longer compiles, `render_chat_prompt` is now a thin wrapper: `ChatPromptRenderer::new(cfg).render(messages)`. Same public API, single source of truth.

Verified locally: `cargo fmt --check`, `cargo build -p cascadia`, `cargo test --workspace --all-targets` all green; `python3 -m py_compile tools/export_gemma4.py` OK. (The Gemma-4-31B export/run validation in the description is unaffected — that commit is unchanged.)

PR #73, opened 2026-06-08
feat(transport): env-configurable activation recv timeout

pawanpaudel93:
Stacked on #70 (`fix/expose-chat-prompt`) — diff is the single commit `ef6eb93`.

Makes the activation `recv` timeout configurable via `CASCADIA_ACTIVATION_TIMEOUT_SECS` (defaults to the prior value when unset), enabling a short per-step timeout for fast-failover under N-stage cold-compile / engine stalls instead of a fixed long block.

**Why:** `labscommunity/cascadia-enterprise` PR [#29 ](https://github.com/labscommunity/cascadia-enterprise/pull/29)(bridge activation splice) pins this runtime to `ef6eb93b` = this commit + #70's `render_chat_prompt`. Merge #70 first, then this; enterprise re-pins to the resulting `main` rev.

**Risk:** additive — falls back to the prior default when the env var is unset; no behavior change for existing callers.


7i7o (approved):
tACK 👍

PR #74, opened 2026-06-10
fix(openvino): rate-limit per-call WARN on failing step()

pawanpaudel93:
## What & why

`OvRuntimeEngine::step()` logged one WARN per failing call (~90k/s when a dropped upstream makes recv fast-error) — the source of cascadia-enterprise#30's 770 GB `chain.log`. Adds a crate-shared `StepWarnLimiter` and wires it into **both** `OvRuntimeEngine` and `Gemma4Engine` (which had the byte-identical flood).

## Related

- **Enterprise half — fixes A + C + consumer:** labscommunity/cascadia-enterprise#31 (relay-loop backoff + log rotation/retention; pins this rev).
- Upstream half (fix B) of labscommunity/cascadia-enterprise#30.

## Changes

- **`warn_limit.rs`** (new, shared): first failure logs in full; later ones counted; 30s `"still failing, N suppressed"` summary; success closes the streak (recovery INFO) and re-latches. ~2 lines/min worst case while wedged.
- **`runtime.rs` + `gemma4.rs`**: `Err` → limiter; `Ok` closes the streak only behind `!v.is_empty() || !self.spec.is_first_stage` (a failed first-stage step clears `active`, so the next idle `Ok(empty)` must not spuriously close it; a relay-stage `Ok` is always a completed round).
- **gemma4 `step_first` inner WARN → debug** (the outer `step()` already rate-limits the same error).
- **`dist_spec` left**: its CPU-speed flood path (socket-closed) is already throttled; the else-branch errors come from a blocking `recv` (frame-rate-bounded, not a spin).

## Test plan

Run from repo root; tick each as it passes.

- [ ] **Limiter + engines** — 28 tests incl. the 3 `warn_limit` state-machine tests (first-failure latch; 30s summary with counter reset; success relatch)
  ```bash
  cargo test -p cascadia-engine-openvino --lib
  ```
- [ ] **No new lint / format**
  ```bash
  cargo clippy -p cascadia-engine-openvino
  cargo fmt -p cascadia-engine-openvino -- --check
  ```

## Review focus

- `StepWarnLimiter` state machine: latch → 30s summary with per-interval counter reset → relatch.
- The idle-first-stage `Ok` gate across first / single (first+last) / middle / last topologies.
- First occurrence still logs the original `"ov-runtime step failed"` / `"gemma4 step failed"` strings, so existing alert matchers keep firing.

## Notes

Logging-cadence change only — `step()` return values are byte-identical, so downstream relay logic is unaffected. The limiter is pure logic with an injected `Instant`, so the state machine is fully unit-tested without an OV runtime. Enterprise consumes this via a pinned git rev (cascadia-enterprise#31).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,6 @@ models/
 *.gguf
 shards/
 shards_cached/
+
+# Claude Code (local agent state + worktrees)
+.claude/
--- a/crates/cascadia-engine-openvino/src/gemma4.rs
+++ b/crates/cascadia-engine-openvino/src/gemma4.rs
@@ -42,7 +42,9 @@ use cascadia_types::{Chunk, GenerationTask, LoadProgress, PeerLayout, ShardSpec,
 use futures::stream;
 use serde::Deserialize;
 use tokenizers::Tokenizer;
-use tracing::{info, warn};
+use tracing::{debug, info, warn};
+
+use crate::warn_limit::{StepWarn, StepWarnLimiter};
 
 // -------- pipeline / stage config --------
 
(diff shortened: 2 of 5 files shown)


7i7o (approved):
tACK 👍

Issue #77, opened 2026-06-11
Add Qwen3.6-35B-A3B (hybrid Gated-DeltaNet MoE) support + upgrade OpenVINO to 2026.2


t8:
## Summary

Two coupled deliverables:

1. **Upgrade OpenVINO GenAI from 2026.1 → 2026.2** across the build, shim, docs, and install tooling.
2. **Add support for Qwen3.6‑35B‑A3B** — Alibaba's hybrid Gated‑DeltaNet + sparse‑MoE model (Apache‑2.0, released 2026‑04‑15).

These are **sequenced, not independent**. OpenVINO **2026.2** (released 2026‑05‑28) adds *native* runtime + GenAI support for the Qwen3.5/Qwen3.6 `qwen3_5_moe` architecture (Gated‑DeltaNet linear attention) on CPU & GPU. That collapses Qwen3.6 from "write a brand‑new exporter and a recurrent‑state pipeline from scratch" down to "bump the SDK and point our existing `ov-genai` `LLMPipeline` at an Optimum‑exported IR." **Do Part A first; it is the enabler for the fast path in Part B.**

---

## Why these two are coupled

| Today (2026.1) | After 2026.2 |
|---|---|
| `qwen3_5_moe` / Gated‑DeltaNet is unknown to the OV runtime — neither `ov-genai` nor `optimum-intel` can compile it. | OV 2026.2 runtime + GenAI compile and run Qwen3.5 **and** Qwen3.6 on CPU & GPU, incl. tool‑calling. |
| `cascadia shard` (multi‑stage exporter) rejects all MoE in `is_moe_config()`. | Unchanged — `cascadia shard` still can't do this arch. But the **single‑stage `ov-genai` engine** can, via the upgraded SDK. |

So the cheapest viable Qwen3.6 path rides entirely on the 2026.2 upgrade + our *already‑built* `ov-genai` engine. The harder pipeline‑parallel path (the real Cascadia value prop) is a separate, deferred phase.

---

# Part A — Upgrade OpenVINO GenAI 2026.1 → 2026.2

**Target SDK:** OpenVINO GenAI **2026.2.0.0** (archive e.g. `openvino_genai_ubuntu24_2026.2.0.0_x86_64.tar.gz`, install root e.g. `/opt/intel/openvino_genai_2026.2.0.0`).

### What 2026.2 brings that matters to us
[...]

PR #78, opened 2026-06-11
chore: add OpenVINO GenAI 2026.2 support

pawanpaudel93:
Part A of labscommunity/cascadia#77: OpenVINO GenAI **2026.2** baseline (the release with native `qwen3_5_moe` support). Base branch for #79.

## Changes
- SDK floor `2026.1+` → `2026.2+` and example paths → `2026.2.0.0` across: `build.rs` panic hint, `setup-openvino.{sh,ps1}`, `INSTALL` / `README` / `QUICKSTART` / `docs/STATUS` / `docs/ARCHITECTURE`, `Dockerfile` `OPENVINO_URL`.
- Removed dead `openvino*` crates.io deps from workspace `Cargo.toml` (absent from `Cargo.lock`; real binding is the C++ shim).
- Added `tools/requirements.txt` — exporter floors (`transformers>=5.2`, `openvino>=2026.2`, `optimum-intel`, `nncf`, `torch`); exact `==` pins after the first verified Qwen3.6 export.

## Deliberately unchanged
- **sparse-MoE token-by-token prefill** (`cascadia-engine-sparse-moe/src/runner.rs:2285`) — workaround for the OV **2026.1.0 CPU "snippets" shape-specialization** bug. 2026.2 release notes list no snippets fix, so we can't confirm it's gone; removing it blind (→ batched `[1,N]` prefill) risks silent corruption. Lifting it touches a different engine + needs an on-CPU parity run — follow-up, not this bump.
- **Dated `2026.1` strings a `grep` still hits are historical facts, not version floors:** the benchmark run-record in `docs/perf/DEVICE_PROFILE.md` (pins the exact build it was measured on) + its PR-34 note, the "tested on miner, OV 2026.1" line in `docs/architectures/gemma4-support.md`, and the workaround comment above. Bumping them would falsify what was actually run.

## Verification
- Stub `cargo check` clean; real `--features openvino` build on the 2026.2 SDK exercised on the Intel fleet via #79.
- `grep '2026\.1'` clean except the dated records above.


7i7o (approved):
tACK 👍

PR #80, opened 2026-06-12
feat(engine,transport,runner): pipeline-link reliability & dead-peer recovery

pawanpaudel93:
Upstream half of enterprise issue [labscommunity/cascadia-enterprise#34](https://github.com/labscommunity/cascadia-enterprise/issues/34). Reliability fixes to the pipeline-parallel engine / transport / runner so a dead or misbehaving peer is **detected and recovered cleanly** instead of wedging, corrupting the next request, or hot-spinning.

## What changed

- **Visible step failures.** `Engine::step` now returns `EngineResult` (was infallible) — a failing step is surfaced, not swallowed into an eternal spin. The error carries the failed `TaskId`, so one task's failure can't terminate an unrelated concurrent task's stream.
- **Real `cancel`** on `OvRuntimeEngine` (was a no-op) — frees the engine slot on abandon instead of draining to `max_tokens`.
- **Frame-start idle ceiling** for black-holed peers: bounds the wait for a frame's first byte (config > env > 900s default; `0` = unbounded), clamped to ≥ the activation recv timeout so it can't re-create a premature kill.
- **Connection-fatal teardown.** A peer crash (TCP RST), a mid-frame stall, or a ceiling fire now drops the socket and is classified fatal — so a late/abandoned frame can never be misread into the next request, and the relay loop **exits for a clean supervisor rebuild** instead of spinning. Covers `ov-runtime`, `dist-spec`, and `sparse-moe`.
- **Relay-loop throttle.** A persistently-failing `step()` backs off instead of pegging a core + flooding logs.

## Reviewer notes

- Connection-fatal classification is string-matched on flattened transport errors (the existing pattern in this repo); the socket is dropped at its owner, and `EngineError::is_connection_fatal()` + the dist-spec worker classifier are kept in sync.
- `recv_exact` uses a whole-buffer wall-clock timeout, not an idle/no-progress one — correct for today's small inter-stage frames (wall-clock ≈ stall); an idle timeout is a noted follow-up if large mid-frame transfers (e.g. future KV blobs) land.
- Dead-peer recovery relies on systemd `Restart=on-failure` (`docs/deploy/cascadia-worker.service` updated to put `StartLimit*` in `[Unit]` so the burst cap applies).

## How to test

No hardware needed. Build the `cascadia` binary first — the `cascadia-tests-e2e` harness spawns it (`target/debug/cascadia`), and a bare `cargo test --workspace` / `just check` does not build bins:

```bash
cargo build -p cascadia
```

```bash
cargo test --workspace
```

Key areas:

```bash
cargo test -p cascadia-transport
```

ceiling-fatal / peer-RST / mid-frame-stall fatal, clamp/resolver units, idle stress.

```bash
cargo test -p cascadia-runner
```

relay dead-peer exit, persistently-failing-step throttle, step-`Err` task attribution.

## Base

Base `pawan/configurable-activation-timeout`. Lands in the runtime pinned rev; enterprise (cascadia-enterprise#37) then bumps the pin to consume it.

### DIFF
--- a/crates/cascadia-api/src/lib.rs
+++ b/crates/cascadia-api/src/lib.rs
@@ -1378,6 +1378,9 @@ fn engine_error_response(err: cascadia_engine::EngineError) -> axum::response::R
         | EngineError::ShardRejected(_) => StatusCode::BAD_REQUEST,
         EngineError::ModelNotFound(_) => StatusCode::NOT_FOUND,
         EngineError::Backend(_) | EngineError::Io(_) => StatusCode::INTERNAL_SERVER_ERROR,
+        // A task-attributed step failure: the engine abandoned a task; the
+        // underlying cause is a backend/transport failure.
+        EngineError::Task { .. } => StatusCode::INTERNAL_SERVER_ERROR,
     };
     (status, Json(serde_json::json!({"error": err.to_string()}))).into_response()
 }
--- a/crates/cascadia-cli/src/lib.rs
+++ b/crates/cascadia-cli/src/lib.rs
@@ -1164,10 +1164,19 @@ async fn cmd_worker(args: WorkerArgs) -> Result<()> {
     if !is_first {
         info!("entering relay loop");
         let r = runner.clone();
-        tokio::task::spawn_blocking(move || r.run_relay_loop())
-            .await
-            .ok();
-        return Ok(());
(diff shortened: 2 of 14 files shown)


7i7o (approved):
tACK 👍

PR #81, opened 2026-06-18
feat(api): OpenAI tool calling (function calling) [cascadia-enterprise#44]

pawanpaudel93:
## OpenAI tool calling — upstream `cascadia-api` · enterprise#44

Adds OpenAI tool calling to the chat API: request/response schema, the shared model-output parser, tools-aware rendering, non-stream + streaming wiring. The **upstream half** (the bulk); the enterprise gateway (#47) consumes it via the engine pin. **Stacked on #74** (`pawan/issue-30-warn-ratelimit`, not `main`). 3 files.

### Review map
- `crates/cascadia-api/src/lib.rs` — all the logic:
  - `ChatCompletionRequest` += `tools`/`tool_choice`; `ChatMessage` += `tool_calls`/`tool_call_id`/`name` + null-tolerant `content` (accepts `tool` role).
  - `parse_tool_calls` + `call_from_xml_function` — three dialects: Hermes JSON `<tool_call>{…}`, Llama `<|python_tag|>`/whole-JSON, Qwen3 `<function=…><parameter=…>` XML. Per-block, never-panics, name-required, `arguments`/`parameters` alias. **Brace-balanced/escape-aware** block extraction (a literal `</tool_call>` inside an argument value can't truncate the block) + **bare whole-output JSON requires args/params** (a data object with a stray `name` isn't misread as a call).
  - Renderer forwards tools + per-message tool fields into the chat template; `build_choice` + streaming gate on `tool_choice` (`none` skips). **Non-tool paths byte-identical.**
- `Cargo.toml` — minijinja `json` feature (so `{{ tool | tojson }}` resolves).
- `Cargo.lock`.

**Verify:** non-tool path byte-identical · parser never panics + requires a name · `tool_choice:"none"` skips · `tojson` filter resolves.

## Test plan (offline — `cargo test -p cascadia-api`)
Repo root, branch `pawan/issue-44-tool-calling`.

- [ ] **Build + lib tests** (36/36; non-tool path byte-identical)
```bash
cargo build -p cascadia-api
cargo test  -p cascadia-api
```
- [ ] **The 3 regression tests the rig surfaced**
```bash
cargo test -p cascadia-api chat_env_supports_tojson_filter parse_qwen3_xml_function_dialect parse_qwen3_xml_multi_param
```

**Live rig:** runs from the enterprise side — cascadia-enterprise#47 pins this tip (`d94ab8d`) and exercises it across 5 engines (commands there, not duplicated). GREEN on all five.

### Why the non-obvious changes
**Rig-surfaced (live probe; unit tests had missed them):**
- **`json` feature:** real Llama/Qwen tool templates serialize via `{{ tool | tojson }}`; without minijinja's `json` feature `tojson` was unregistered → render threw `unknown filter: tojson` → silent fallback to the legacy formatter dropped tools (the model wrote Python). Test: `chat_env_supports_tojson_filter`.
- **Qwen3 XML parser:** the MoE emits `<function=get_weather><parameter=city>Paris</parameter></function>`, not JSON Hermes → was dropped → `finish=stop`. Added `call_from_xml_function` + 2 tests.

**Review/stress-surfaced (multi-agent review of the diff):**
- **Brace-balanced block scan (M2):** the JSON `<tool_call>` block was delimited by the first `</tool_call>` substring → a literal `</tool_call>` inside an argument value truncated it → silent drop. Now a string/escape-aware brace scan extracts the object; close-tag fallback for the XML dialect. Tests: `parse_tool_call_arg_containing_close_tag`, `parse_tool_calls_brace_balance_stress`.
- **Bare-JSON tool-shape guard (M3b):** a whole-output JSON with no `<|python_tag|>` was treated as a call on a bare `name` key → a plain data answer `{"name":"Alice",…}` became a phantom call. Now the bare path requires `arguments`/`parameters` (real Llama calls always carry them); `<|python_tag|>` + `<tool_call>` paths stay lenient. Test: `parse_bare_json_requires_tool_shape`.

### Follow-up
Constrained-decoding / structured-output **enforcement** on the embedded engines (`required`/`strict`/`response_format` hard-guarantee — most of the engine work is here in `cascadia`) is out of scope and tracked, research-backed + staged, in **labscommunity/cascadia-enterprise#49**.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1669,6 +1669,7 @@ checksum = "805bfd7352166bae857ee569628b52bcd85a1cecf7810861ebceb1686b72b75d"
 dependencies = [
  "memo-map",
  "serde",
+ "serde_json",
 ]
 
 [[package]]
--- a/crates/cascadia-api/Cargo.toml
+++ b/crates/cascadia-api/Cargo.toml
@@ -27,7 +27,10 @@ bytes = { workspace = true }
 # `{% macro %}` helpers for tool/role formatting); without it the template
 # fails to parse and rendering silently falls back to the legacy formatter,
 # which degenerates instruct models. `multi_template` covers `{% include %}`.
-minijinja = { version = "2", default-features = false, features = ["builtins", "serde", "macros", "multi_template"] }
+# `json` registers the `tojson` filter that Llama-3.1/Qwen tool templates use to
+# serialize tool schemas; without it tool requests error and drop to the legacy
+# formatter, silently stripping tools from the prompt.
+minijinja = { version = "2", default-features = false, features = ["builtins", "serde", "macros", "multi_template", "json"] }
 minijinja-contrib = { version = "2", default-features = false, features = ["pycompat"] }
 
 [dev-dependencies]
--- a/crates/cascadia-api/src/lib.rs
+++ b/crates/cascadia-api/src/lib.rs
@@ -2424,6 +2424,50 @@ mod tests {
     }
 
     #[test]
+    fn parse_tool_calls_brace_balance_stress() {
+        // 1) nested object/array args + a '}' and a '{' inside a string value +
+        //    an escaped quote — exact JSON must survive the balanced scan.
+        let calls = parse_tool_calls(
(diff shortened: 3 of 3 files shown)


7i7o (approved):
tACK 👍

PR #85, opened 2026-06-21
feat(sparse-moe): pipeline-parallel MiniMax-M2 across ranks

t8:
## What

Lets the MiniMax-M2 OV-IR shell backend shard across ranks (pipeline-parallel), instead of single-stage only. Stacked on #65 (base `feat/minimax-m2`); retarget to `main` once #65 merges.

## How

Wires M2 into the **same** pipeline-parallel framework the K2.6 path already uses — the `dist` wire protocol (Forward/Reset/Token, F32 hidden `[1,1,H]`) is engine-agnostic, so M2 (H=3072) needed **zero** wire changes.

- `OvMoeRunner::load_staged(rank, total)`: each rank loads a contiguous even slice of the transformer shells (+ their experts); rank 0 also loads the embedding (`layer0`), the last rank also loads the head. `embed`/`head` become `Option`. `step_logits` is split into `embed_token` / `forward_layers(hidden, pos)` / `head_logits` so the whole-model runner and a per-rank slice run identical math. `load()` is now `load_staged(.., 0, 1)` — single-stage behavior unchanged.
- `OvMoeEngine` gains a multi-stage path mirroring `SparseMoEEngine`: rank 0 drives (embed → its layers → ship hidden downstream → await the sampled token), middle ranks relay, the last rank runs head + sampler and returns the token upstream. No spec-decode (M2 sends only per-token Forward frames). The Builder loads the per-rank slice, gates the tokenizer to rank 0, and passes transport/rank/total into the engine.

## Validation (miner, OpenVINO 2026.1)

- **Exact correctness** — new `minimax_m2_two_rank_pipeline_matches_hf_reference`: the tiny synthetic M2 split across 2 ranks reproduces the canonical HF greedy stream exactly (`[1,2,3,4,5,231,231,229,231,229,231]`). The single-stage tiny gate still passes (no regression).
- **Real distributed engine over TCP (tiny)** — two `cascadia worker` processes connect over sockets and load the correct slices (rank0 = embed+`0..0`, rank1 = `1..1`+head).
- **Real distributed engine over TCP (full 230B)** — the sharded model, served through the OpenAI API across 2 ranks, answered *"What is the capital of France?"* → `"answer": "Paris"` (12 tokens, ~0.087 tok/s). Exercises the whole `step_first` driver → `dist` frames → `step_worker`/`handle_forward` → last-rank head+sampler → Token-upstream → decode path end-to-end with the real model.

Loopback uses the identical transport code as a cross-host deployment (127.0.0.1 vs a LAN IP).

## Known limitation (v1, matches the K2.6 path)

The last rank's repetition-penalty `history` includes prefill-stage samples (it samples every Forward). Greedy (`repetition_penalty == 1.0`) is unaffected, so a greedy pipeline run matches single-stage exactly. Documented in the `OvMoeEngine` doc + `dist.rs`.

## Test

```bash
M2_MODEL_DIR=/path/to/tiny_m2 cargo test -p cascadia-engine-sparse-moe \
  --features openvino --test minimax_m2_eval -- --nocapture
```

### DIFF
--- a/crates/cascadia-engine-sparse-moe/src/engine.rs
+++ b/crates/cascadia-engine-sparse-moe/src/engine.rs
@@ -349,19 +349,19 @@ impl Builder for SparseMoEBuilder {
         }
 
         // OV-IR shell backend (MiniMax-M2): the whole architecture lives in
-        // the traced graphs, so we run a dedicated single-stage runner
-        // instead of the K2.6 MLA kernel. Pipeline-parallel isn't wired for
-        // this backend yet (the K2.6 transport path assumes the Rust shell).
+        // the traced graphs, so we run a dedicated runner instead of the
+        // K2.6 MLA kernel. Single-stage loads the whole model; multi-stage
+        // (total > 1) loads a contiguous layer slice per rank and exchanges
+        // F32 hidden states over the same engine-agnostic dist wire protocol
+        // the K2.6 path uses (rank 0 embeds + drives, the last rank runs the
+        // head + sampler).
         let is_ov = Manifest::load(&self.config.model_dir)
             .map(|m| m.is_ov_shell())
             .unwrap_or(false);
         if is_ov {
-            if self.config.total > 1 {
-                return Err(EngineError::InvalidConfig(
-                    "ov_ir shell backend (MiniMax-M2) is single-stage only; total must be 1".into(),
-                ));
-            }
             let cfg = self.config.clone();
+            let total = cfg.total.max(1);
+            let rank = cfg.rank.min(total - 1);
             let opts = resolve_runner_options(&cfg);
             let cap = opts.max_cached_experts;
             let plugin_for_worker = plugin.clone();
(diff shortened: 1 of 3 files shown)


t8:
## Cross-host validation (miner + NUC) ✅

Ran the sharded pipeline across two physical machines, plus a follow-up fix.

**Cross-host, both CPU — byte-exact:**

| Run | rank0 (embed+layer0) | rank1 (layer1+head) | Output |
|---|---|---|---|
| Loopback | miner CPU (Linux) | miner CPU (Linux) | `t231t61t29t235t162t29` |
| **Cross-host** | **miner CPU (Linux)** | **alpha NUC CPU (Windows)** | **`t231t61t29t235t162t29`** |

rank0 on the miner (Linux x86) connected to rank1 on a Windows NUC over tailscale; the hidden states crossed the network and the distributed result is **numerically identical** to the single-host run — i.e. the cross-host transport is correct. Tiny synthetic model (the full 230B can't run on a NUC: its `int4_bin` experts need AVX-512, which consumer Lunar/Panther Lake lack).

**Follow-up fix (34c5fff):** running a rank on the NUC's Intel iGPU surfaced that `OvMoeRunner` set `SNIPPETS_MODE=DISABLE` unconditionally — a **CPU-plugin** option the GPU plugin rejects (`Option not found: SNIPPETS_MODE`). Gated it to `device == CPU`. With the fix the iGPU rank now loads; the forward itself then hits an OV GPU-plugin/driver error (`[GPU] CL_OUT_OF_RESOURCES`) executing M2's custom shell graph on the Panther Lake iGPU — a plugin/device limit (M2 is a CPU/high-RAM model, so the iGPU isn't its target), handled gracefully by the worker.

Net: pipeline-parallel M2 is validated correct in-process (exact HF parity), over real TCP sockets single-host (full 230B → "Paris"), and across two physical machines (byte-exact).

PR #86, opened 2026-06-21
feat(sparse-moe): run MiniMax-M2 on Intel iGPUs via router-split

t8:
## What

Lets MiniMax-M2 shells execute on Intel iGPUs (Arc / Xe), so NUC iGPUs can be pipeline stages. Stacked on #85 (pipeline-parallel); retarget to `main` after #85.

## Why

The M2 shell's **MoE router** subgraph (gate MatMul → sigmoid → +`e_score_correction_bias` → TopK → Gather → renorm) makes the OpenVINO Intel GPU plugin emit a kernel that fails at runtime with `CL_OUT_OF_RESOURCES` — and can wedge the OpenCL context until a cold reboot. This reproduces on OpenVINO **2026.1.0, 2026.2.1 (latest stable), and 2026.3.0-nightly**, and is an open upstream bug (openvinotoolkit/openvino#35723, plus #25141/#28919/#32665 closed without a fix). Diagnosis (runtime-graph dump + IR bisection) showed the **attention/KV core, embed, experts, and head all run on the iGPU fine** — only the router fails.

## How

Carve each shell into two graphs:
- `shell_core.xml`: `x, past_k, past_v, past_seq_len → attn_out_post_norm, attn_residual, shared_expert_out, present_k, present_v` (runs on the iGPU)
- `router.xml`: `apn_in → routing_ids, routing_weights` (runs on the CPU plugin; trivial compute)

`tools/split_m2_shells.py` post-processes an export (OV graph surgery: replace the `attn_out_post_norm` tensor with an `apn_in` Parameter for the router; output-prune the router from the core). Validated byte-exact vs the monolithic shell.

`OvMoeRunner`: when the target device is not CPU **and** `shell_core.xml` exists, it compiles `shell_core` on the device and `router` on the CPU plugin; `forward_layers` reads the 5 core outputs from the shell and runs the router separately for ids/weights. On CPU (or without split files) the monolithic shell path is unchanged. `ShellPorts` → `CorePorts` + `RouterPorts`; manifest gains `shell_core_xml`/`router_xml`. Also `--tiny-layers` to size the synthetic model for N-rank pipeline tests.

## Validation (Panther Lake NUCs, Arc B390)

- **Python e2e on a NUC iGPU**: tiny M2 split (core+experts+head on GPU, router on CPU) reproduces the canonical HF greedy stream **exactly**.
- **Real cascadia binary, single-stage on a NUC iGPU**: log shows `compiled N shells on GPU (router_split=true)`; output matches the CPU reference.
- **3-node pipeline**: miner **CPU** (rank0, embed+layers) + two NUC **iGPUs** (rank1, rank2+head), 4-layer tiny model, over the network — generated end-to-end; both NUC logs confirm `compiled shells on GPU (router_split=true)`.

## Scope note

Only small models fit the NUCs; the full 230B M2 still can't run there (`int4_bin` experts need AVX-512; 16 GB iGPU; RAM). This is a small-model / multi-node-iGPU capability, not a path for production-scale M2.

## Test

```bash
python tools/split_m2_shells.py --model-dir /path/to/m2_export
cascadia worker --engine sparse-moe --model /path/to/m2_export --device GPU ...
```

### DIFF
--- a/crates/cascadia-cli/src/lib.rs
+++ b/crates/cascadia-cli/src/lib.rs
@@ -221,6 +221,19 @@ pub struct WorkerArgs {
     #[arg(long)]
     pub total: u32,
 
+    /// First transformer layer this stage holds (global, 0-based, inclusive).
+    /// Together with --layer-end, pins an explicit asymmetric layer split
+    /// instead of the default even split across ranks — e.g. a high-RAM CPU
+    /// node holding most layers while small iGPU nodes hold a few each.
+    /// `0/0` (both default) = even split. MiniMax-M2 sparse-moe only.
+    #[arg(long, default_value_t = 0)]
+    pub layer_start: u32,
+
+    /// One-past-the-last transformer layer this stage holds (global,
+    /// exclusive). `0` = unset (even split). See --layer-start.
+    #[arg(long, default_value_t = 0)]
+    pub layer_end: u32,
+
     /// HF model id or local model directory.
     #[arg(long)]
     pub model: String,
@@ -457,6 +470,8 @@ impl WorkerArgs {
         WorkerArgs {
             rank: 0,
             total: 1,
+            layer_start: 0,
+            layer_end: 0,
             model,
             listen: ":9100".into(),
             next: None,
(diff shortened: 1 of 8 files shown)


t8:
Auto-closed when its base branch `feat/minimax-m2-pipeline-parallel` was deleted on merge of #85. Re-opened as #87 (same head branch, retargeted to `main`).

Issue #55, opened 2026-06-01
ov-genai: /v1/chat/completions usage token counts are always 0


t8:
## Summary

The OpenAI-compatible `/v1/chat/completions` response from the **ov-genai** engine reports `usage` token counts as all zeros, even though the engine internally knows and logs the real generated-token count. Clients that rely on `usage` for billing, rate-limiting, or context-budget accounting get no signal.

## How it was found

On-hardware validation on an Intel AI PC (Panther Lake, OpenVINO 2026.1), `cascadia run` of `OpenVINO/TinyLlama-1.1B-Chat-v1.0-int4-ov`. The response body:

```json
{"choices":[{"index":0,"message":{"role":"assistant","content":"The capital of France is Paris."},"finish_reason":"stop"}],
 "usage":{"prompt_tokens":0,"completion_tokens":0,"total_tokens":0}}
```

…while the engine log for the *same* request shows the real count:

```
INFO cascadia_engine_openvino::genai: task done task=chatcmpl-... tokens=8 elapsed_s=0.2339 tok_s=34.2
```

So the count exists at the engine layer (`tokens=8`) but isn't propagated into the API `usage` object.

## Expected

`usage.completion_tokens` (and ideally `prompt_tokens` / `total_tokens`) in the chat-completions and completions responses should reflect the engine's actual token accounting for the ov-genai path. The engine already surfaces the generated-token count; wire it through the runner/chunk stream into the API response builder.

## Acceptance

- A non-streaming `/v1/chat/completions` response has `completion_tokens` matching the number of tokens generated (the value the engine logs).
- `prompt_tokens` populated where the tokenizer makes it available; `total_tokens = prompt_tokens + completion_tokens`.
- Streaming responses carry usage in the final chunk if `stream_options.include_usage` is set (OpenAI-compatible behavior), or this is explicitly documented as out of scope.

Issue #20, opened 2026-05-18
Continuous batching for ov-genai via ContinuousBatchingPipeline


t8:
## Summary

OV-GenAI's continuous batching is default on the GPU plugin since OpenVINO 2025.1. Tahoma's `ov-genai` engine runs one task at a time — `step()` pops a single task from `pending` and runs the entire generation to completion before returning (`crates/tahoma-engine-openvino/src/genai.rs:240-285`). The HTTP API gates concurrent clients behind a `tokio::sync::Semaphore` returning 503 (`crates/tahoma-api/src/lib.rs:369-380`).

Restructure `OvGenaiEngine` to feed multiple `pending` tasks into `LLMPipeline` simultaneously, configure `scheduler_config`, route outputs back to each task's `Chunk` stream by task ID. Removes the need for HTTP 503 backpressure at moderate concurrency.

## Why

OV-GenAI 2026.1's `LLMPipeline` continuous-batches with paged attention by default on GPU. This gives:

- Up to 55% KV-cache memory reduction (per OVMS benchmark)
- Up to 2.2× throughput at moderate concurrency
- Prefix caching across requests (free)

Tahoma's current `ov-genai` engine throws this away — it serializes requests behind a mutex and a semaphore. For single-user inference this doesn't matter; for any multi-user deployment it's the single biggest LLM-perf knob Tahoma is leaving on the table.

This is the prerequisite for cross-host continuous batching across PP stages (separate RFC issue) — get single-host CB right first.

## Current state

`crates/tahoma-engine-openvino/src/genai.rs`:

- `OvGenaiBuilder::with_*` configures the pipeline at build time
- `OvGenaiEngine` holds an `LLMPipeline` instance
- `submit(task)` pushes to `pending: VecDeque<GenerationTask>`
- `step()` pops one task, calls `pipeline.generate(prompt, generation_config)` synchronously, builds a single `Chunk` with the entire output, returns

OV-GenAI continuous batching requires using `LLMPipeline::generate` with **multiple concurrent calls** or the `ContinuousBatchingPipeline` API. The latter is cleaner for this use case.

### (1) Switch to ContinuousBatchingPipeline

**File:** `crates/tahoma-ov-genai-shim/cpp/shim.{h,cpp}`

Add FFI entry points for `ov::genai::ContinuousBatchingPipeline`:

```c
extern "C" void* tahoma_cb_pipeline_create(
    const char* model_path,
    const char* device,
[...]

PR #91, opened 2026-06-24
feat(api): OpenAI sampling params + finish_reason + streaming usage (#14)

### LINKED ISSUE #14: Surface sampling parameters in /v1/chat/completions (top_p, top_k, seed, stop, finish_reason, logprobs)

t8:
Closes #14.

> **Stacks on #90 (#55).** This branch is cut from `fix/ovgenai-usage-tokens-55`, so it contains that commit too (`b4a58bc`). Merge #90 first, or this will be rebased onto main after it lands. The #14 work is the second commit (`21c15f2`).

Surfaces the standard OpenAI sampling + observability fields on `/v1/chat/completions` and routes them through the engines.

## What's in
| Area | Change |
|---|---|
| **Request** | `top_p`, `top_k`, `seed`, `stop` (string\|array), `frequency_penalty`, `presence_penalty`, `logprobs`, `top_logprobs`, `stream_options` — bundled into `cascadia_types::SamplingParams` on `GenerationTask`. |
| **finish_reason** | New `FinishReason` enum on the final `Chunk`; API emits `"length"` when the engine hit `max_tokens`, else `"stop"` (was hardcoded `"stop"`). |
| **Sampler (sparse-moe)** | `SamplingConfig` + `sample()` gain `top_k`, `frequency_penalty`, `presence_penalty`. |
| **ov-genai** | Forwards `top_p`/`top_k`/`seed`/penalties into `ov::genai::GenerationConfig` via new shim setters. |
| **Multi-stage wire** | `SAMPLING_WIRE_BYTES` 28→40; `Forward`/`ForwardBatch` codes bumped `0x02→0x04` / `0x03→0x05` so a peer on the old 28-byte layout fails fast at `parse_kind` (the issue's suggested `0x03` was already taken by `ForwardBatch`). |
| **Streaming** | `stream_options.include_usage` emits a final `usage` SSE chunk. |
| **stop** | Stop sequences truncate the assembled output at the earliest match (API-level). |
| **logprobs** | Request + OpenAI response schema (`ChatLogprobs`) wired; surfaced when chunks carry per-token logprobs. |

## Validation status (honest)
**Unit/integration tested (stub build, no hardware):**
- ✅ Sampler: top_k forces argmax under temperature; frequency penalty count-scales; presence penalty is count-independent.
- ✅ Wire: 40-byte round-trip incl. new fields; frame-code bump + old-code rejection.
- ✅ API: `finish_reason` length-vs-stop, stop-sequence truncation, full-param-set acceptance (no 400), streaming `usage` frame.
- ✅ `cargo fmt --all --check` clean; full workspace test green; no regressions.

**Needs on-HW validation before merge (no OpenVINO/Intel on this dev box):**
- ⏳ ov-genai actually honoring the new `GenerationConfig` knobs — the C++ shim setters are a mechanical mirror of the existing ones over standard OV fields, and the **AI-PC CI tier compiles the openvino path**, but live behavior (e.g. `seed` reproducibility, `top_p` effect) should be confirmed on a real model.
- ⏳ 2-box multi-stage sparse-moe carrying the new sampling fields end-to-end (wire round-trips in-process; a real 2-rank run confirms the bump).

## Deferred follow-ups (called out, not silently dropped)
- **logprobs emission per engine.** Request + response schema are in place, but no engine emits per-token logprobs yet: ov-genai needs `DecodedResults.scores` extraction through the shim, and sparse-moe single-stage emits one final chunk so it needs per-token emission (or a `Vec<TokenLogprobs>` on the final chunk). `logprobs: true` is accepted and returns `null` until then.
- **finish_reason for ov-runtime / ov-dist-spec / gemma4** (currently fall back to `"stop"`, unchanged).
- **Engine-side early-stop on `stop` strings** (would also save compute; today truncation is API-level output-only).
- **Streaming stop-sequence trimming** across chunk boundaries.

### DIFF
--- a/crates/cascadia-api/src/lib.rs
+++ b/crates/cascadia-api/src/lib.rs
@@ -296,6 +296,21 @@ pub struct ChatMessage {
     pub content: String,
 }
 
+/// OpenAI `stop` is either a single string or an array of strings.
+#[derive(Deserialize, Debug)]
+#[serde(untagged)]
+pub enum StopSpec {
+    Single(String),
+    Multiple(Vec<String>),
+}
+
+/// OpenAI `stream_options`. Only `include_usage` is honored today.
+#[derive(Deserialize, Debug, Default)]
+pub struct StreamOptions {
+    #[serde(default)]
+    pub include_usage: bool,
+}
+
 #[derive(Deserialize, Debug)]
 pub struct ChatCompletionRequest {
     pub model: String,
@@ -305,12 +320,64 @@ pub struct ChatCompletionRequest {
     #[serde(default)]
     pub temperature: f32,
     #[serde(default)]
+    pub top_p: Option<f32>,
+    #[serde(default)]
+    pub top_k: Option<u32>,
+    #[serde(default)]
+    pub seed: Option<u64>,
+    #[serde(default)]
+    pub stop: Option<StopSpec>,
+    #[serde(default)]
+    pub frequency_penalty: Option<f32>,
+    #[serde(default)]
+    pub presence_penalty: Option<f32>,
+    /// OpenAI `logprobs` (bool). When true, per-token logprobs are returned
+    /// (engine-permitting).
+    #[serde(default)]
+    pub logprobs: bool,
+    /// OpenAI `top_logprobs` (0..=20): how many alternatives per position.
+    #[serde(default)]
+    pub top_logprobs: Option<u32>,
+    #[serde(default)]
     pub stream: bool,
+    #[serde(default)]
+    pub stream_options: Option<StreamOptions>,
     /// Hybrid-reasoning switch (Qwen3+ convention). Default true =
     /// model-default behavior; false asks the engine to skip the
     /// <think> block (engines that can't, ignore it).
     #[serde(default = "default_true")]
     pub enable_thinking: bool,
+}
+
+impl ChatCompletionRequest {
+    /// Build the engine-facing `SamplingParams`, applying OpenAI defaults
+    /// (top_p=1.0, penalties=0.0) when a field is omitted.
+    fn sampling_params(&self) -> cascadia_types::SamplingParams {
+        cascadia_types::SamplingParams {
+            top_p: self.top_p.unwrap_or(1.0),
+            top_k: self.top_k.unwrap_or(0),
+            seed: self.seed,
+            frequency_penalty: self.frequency_penalty.unwrap_or(0.0),
+            presence_penalty: self.presence_penalty.unwrap_or(0.0),
+            stop: self
+                .stop
+                .as_ref()
+                .map(|s| match s {
+                    StopSpec::Single(x) => vec![x.clone()],
+                    StopSpec::Multiple(v) => v.clone(),
+                })
+                .unwrap_or_default(),
+        }
+    }
+
+    /// Top-logprobs count to request from the engine (0 = disabled).
+    fn logprobs_count(&self) -> u32 {
+        if self.logprobs {
+            self.top_logprobs.unwrap_or(1).clamp(1, 20)
+        } else {
+            0
+        }
+    }
 }
 
 fn default_true() -> bool {
@@ -327,12 +394,21 @@ struct ChatChoiceMessage {
     content: String,
 }
 
+/// OpenAI chat `logprobs` object: one entry per generated token under
+/// `content`. Present only when the request set `logprobs: true` AND the
+/// engine emitted per-token logprobs.
+#[derive(Serialize)]
+struct ChatLogprobs {
+    content: Vec<cascadia_types::TokenLogprobs>,
+}
+
 #[derive(Serialize)]
 struct ChatChoice {
     index: u32,
     message: ChatChoiceMessage,
     finish_reason: &'static str,
-    logprobs: Option<()>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    logprobs: Option<ChatLogprobs>,
 }
 
 #[derive(Serialize)]
@@ -578,7 +654,8 @@ async fn chat_completions(
         prompt,
         max_tokens: req.max_tokens,
         temperature: req.temperature,
-        logprobs: 0,
+        logprobs: req.logprobs_count(),
+        sampling: req.sampling_params(),
         enable_thinking: req.enable_thinking,
         trust_remote_code: false,
     };
@@ -608,7 +685,12 @@ async fn chat_completions(
     let inflight = InFlightGuard::new(state.stats.clone());
 
     if req.stream {
-        return stream_completion(state, req.model, task, permit, inflight)
+        let include_usage = req
+            .stream_options
+            .as_ref()
+            .map(|o| o.include_usage)
+            .unwrap_or(false);
+        return stream_completion(state, req.model, task, permit, inflight, include_usage)
             .await
             .into_response();
     }
@@ -624,6 +706,10 @@ async fn chat_completions(
     let mut buf = String::new();
     let mut completion_tokens: u32 = 0;
     let mut prompt_tokens: u32 = 0;
+    // OpenAI `finish_reason`: `length` when the engine hit max_tokens, else
+    // `stop`. Engines that don't distinguish leave it None → "stop".
+    let mut finish_reason: &'static str = "stop";
+    let mut logprobs_content: Vec<cascadia_types::TokenLogprobs> = Vec::new();
     while let Some(chunk) = chunk_stream.next().await {
         // A failed task carries `error` on its final chunk. Without this
         // branch the empty text below would build a normal 200 with empty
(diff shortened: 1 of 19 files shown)


t8:
Addressed the two review NITs in 5a5155b:
- **NIT-1 (corrupt zero `repetition_penalty`):** `decode_sampling` now clamps `<= 0`/NaN back to the 1.0 no-op (min = `MIN_POSITIVE`), so an out-of-version peer can't feed a 0 divisor into the sampler (`logit / α = +inf`). Added a regression test `decode_sanitizes_corrupt_zero_repetition_penalty`.
- **NIT-2 (penalty scope):** documented in-code that frequency/presence penalties deliberately span the full history (prompt + completion), matching the existing `repetition_penalty` behavior — a conscious divergence from OpenAI's completion-only scope.

The AI-PC CI job is the remaining gate (it compiles the C++ shim + exercises ov-genai); a reviewer separately verified all five new `GenerationConfig` setters map to real public members of the targeted OV 2026.2.0.0 header.

t8:
## On-hardware validation (alpha Panther Lake + charlie Lunar Lake, real OpenVINO 2026.1)

Built `--features openvino` on the NUCs and exercised the live engines. Results:

**ov-genai (CPU, TinyLlama int4-ov):** 7/8 ✅ — `top_p`/`top_k`/penalties accepted (the C++ shim setters work on real OV, no crash), `finish_reason=length`, stop truncation, streaming `usage` frame all pass. The 1 fail ↓.
**ov-runtime (iGPU, Llama-3.2-1B shard):** ✅ — sampling accepted; **`seed` reproducible** (Rust sampler).
**sparse-moe 2-stage pipeline (loopback, m2_tiny):** ✅ — a request carrying the extended sampling block (`top_k`/`frequency_penalty`/`seed`) completes across the **bumped 40-byte Forward frame** end-to-end.

### Finding: ov-genai `seed` not reproducible across requests
Two `seed=42` requests on ov-genai → different output; on ov-runtime → identical. **Wiring is correct** — root cause is upstream: OV-GenAI's `StatefulLLMPipeline::generate` only re-seeds when the seed *value changes* between consecutive calls (`if (m_sampler.get_seed() != config.rng_seed) set_seed(...)`) and never clears the per-request RNG context, so repeated identical seeds reuse the advanced RNG. `seed` is still forwarded (different seeds change output), and the Rust-sampler engines honor it fully.

Documented at the wiring site in e525485. Full per-request reproducible ov-genai needs the `ContinuousBatchingPipeline` (fresh request context per request) — i.e. **#20**, not a #14-scope change.

PR #92, opened 2026-06-24
feat(api): add /v1/completions (OpenAI legacy completions) (#19)

### LINKED ISSUE #19: Add /v1/completions endpoint (OpenAI-compatible legacy completions)
## Summary

Add the OpenAI-compatible `/v1/completions` endpoint alongside the existing `/v1/chat/completions`. Many tools (older clients, evals like lm-eval-harness, OpenAI-compat libraries) still use the legacy completions endpoint.

The endpoint takes a raw `prompt` string (or array of strings) and returns a continuation. No chat templating.

## Why

Today `crates/tahoma-api/src/lib.rs:143-149` registers only `/v1/chat/completions`, `/v1/models`, `/health`, `/v1/cancel/:task_id`. Several common tooling paths assume `/v1/completions` exists — notably:

- `lm-evaluation-harness`'s `openai_completions` task adapter
- LangChain's `OpenAI` LLM class (vs `ChatOpenAI`)
- Direct `curl` testing without chat templates
- Any tooling that pre-tokenized the chat template itself (e.g., agentic frameworks measuring template effects)

Implementation is straightforward because the engines already accept a tokenized prompt as `GenerationTask.prompt` — the routing logic just needs to skip the Jinja chat-template render (`tahoma-api/src/lib.rs::render_chat_template`, added in commit `339091a`).

### (1) Define request + response types

**File:** `crates/tahoma-api/src/lib.rs`

```rust
#[derive(Deserialize)]
pub struct CompletionRequest {
    pub model: String,
    pub prompt: PromptSpec,            // string or array of strings (OAI-style)
    #[serde(default)]
    pub max_tokens: Option<u32>,
    #[serde(default)]
    pub temperature: Option<f32>,
    // ... reuse the sampling fields from issue #14 (top_p / top_k / seed / stop / penalties / logprobs / stream)
    #[serde(default)]
    pub stream: Option<bool>,
    #[serde(default)]
    pub echo: Option<bool>,             // prepend prompt to output
    #[serde(default)]
    pub suffix: Option<String>,          // punt as separate issue
}

#[derive(Deserialize)]
#[serde(untagged)]
pub enum PromptSpec {
    Single(String),
    Multiple(Vec<String>),
}

#[derive(Serialize)]
pub struct CompletionResponse {
    pub id: String,
    pub object: String,                  // "text_completion"
    pub created: u64,
    pub model: String,
    pub choices: Vec<CompletionChoice>,
    pub usage: Usage,
}

#[derive(Serialize)]
pub struct CompletionChoice {
    pub text: String,
    pub index: u32,
    pub finish_reason: Option<String>,
    pub logprobs: Option<TokenLogprobs>,
}
```

### (2) Register the route

**File:** `crates/tahoma-api/src/lib.rs` near line 143-149

```rust
.route("/v1/completions", post(completions_handler))
```

### (3) Implement the handler

Reuse most of the chat handler. Differences:
- No `messages` → no Jinja render. Just tokenize `prompt`.
- If `prompt` is `Vec<String>`, each is a separate task; return a `choices` array with `index` matching input position. Punt to v2 — return a 400 for multi-prompt initially.
- `echo: true` means prepend the prompt to the output (rarely used but spec-mandated).

### (4) Multi-prompt batching

OpenAI's spec allows `prompt: ["abc", "def"]` to submit multiple completions in one call. **Out of scope for v1** — Tahoma's engines run one task at a time per the existing `step()` loop. Document the limitation; return a clean 400 with a helpful error when an array is passed.

```rust
if let PromptSpec::Multiple(_) = req.prompt {
    return Err(ApiError::NotImplemented(
        "multi-prompt batching not yet supported; submit one prompt per request"
    ));
}
```

Wire this up properly when [continuous-batching across PP stages](#) or [continuous batching for ov-genai](#) lands.

### (5) Echo handling

OAI's `echo: true` flag prepends the prompt to the output. Mechanically, the API decodes the prompt's `input_ids` back to text and concatenates with the generated text. Trivial. Default `false`.

## Acceptance criteria

- [ ] `POST /v1/completions` with `{"model":"k26","prompt":"The capital of France is","max_tokens":4,"temperature":0}` returns `text: " Paris..."` and `finish_reason: "stop"` or `"length"`
- [ ] Streaming works (`"stream": true` → SSE chunks with `text` deltas; final chunk has `finish_reason`)
- [ ] `echo: true` prepends the prompt to the output (verified by smoke test)
- [ ] Multi-prompt array is documented as not-yet-supported and returns a clean 400 with a helpful error
- [ ] All sampling parameters from #14 are honored when passed
- [ ] Existing `/v1/chat/completions` unaffected

## Effort

~ 0.5–1 day after #14 (sampling parameters) lands. Mostly a fork of the chat handler.

## Out of scope

- Multi-prompt array batching (depends on cross-task batching infra)
- `suffix` field (OpenAI's prefix+suffix completion) — separate small issue
- `best_of` / `n > 1` — separate
- Token-level metadata beyond `logprobs` — covered by #14

## References

- [OpenAI Legacy Completions API reference](https://platform.openai.com/docs/api-reference/completions/create)
- Tahoma Jinja chat template: commit `339091a` on `main`
- Existing route table: `crates/tahoma-api/src/lib.rs:143-149`

## Related

- Depends on #14 (sampling parameters) — `top_p`/`top_k`/`seed`/`stop`/`logprobs`/etc. will be defined there

t8:
Closes #19.

> **Stacks on #91 (#14) → #90 (#55).** Cut from `feat/sampling-params-14`, so it contains those commits too. Merge order: #90 → #91 → this. The #19 change is the top commit (`8989b4e`), a single-file addition to `crates/cascadia-api/src/lib.rs`.

Adds the OpenAI-compatible legacy `/v1/completions` endpoint — a raw-`prompt` sibling of `/v1/chat/completions`. No chat-template render; returns `text` instead of a `message`. Reuses #14's sampling / `finish_reason` / usage / stop-sequence / streaming-usage machinery.

## Behavior
- **Single prompt only.** A `prompt` array returns a clean `400` ("multi-prompt batching not yet supported") — the engines run one task per step; batching is future work.
- **`echo: true`** prepends the raw prompt to the returned text (non-streaming, and as the first SSE frame when streaming).
- **Streaming** emits `text_completion` chunks with `choices[].text` deltas + an optional final `usage` frame (`stream_options.include_usage`).
- **`logprobs`** (legacy integer count) threads through to the task; per-engine emission is the same deferred follow-up as #14.

## Tests (stub build, mock engine)
- ✅ `text_completion` object shape + `usage.completion_tokens`
- ✅ `echo` prepends the prompt
- ✅ multi-prompt array → 400
- ✅ stop-sequence truncation + `finish_reason`
- ✅ streaming `text_completion` deltas + `usage` + `[DONE]`

`cargo fmt` clean; `cargo test -p cascadia-api` green (20 tests). No engine-path changes, so no hardware needed for this PR specifically (it rides on whatever the underlying engine produces).

### DIFF
--- a/crates/cascadia-api/src/lib.rs
+++ b/crates/cascadia-api/src/lib.rs
@@ -296,6 +296,21 @@ pub struct ChatMessage {
     pub content: String,
 }
 
+/// OpenAI `stop` is either a single string or an array of strings.
+#[derive(Deserialize, Debug)]
+#[serde(untagged)]
+pub enum StopSpec {
+    Single(String),
+    Multiple(Vec<String>),
+}
+
+/// OpenAI `stream_options`. Only `include_usage` is honored today.
+#[derive(Deserialize, Debug, Default)]
+pub struct StreamOptions {
+    #[serde(default)]
+    pub include_usage: bool,
+}
+
 #[derive(Deserialize, Debug)]
 pub struct ChatCompletionRequest {
     pub model: String,
@@ -305,12 +320,64 @@ pub struct ChatCompletionRequest {
     #[serde(default)]
     pub temperature: f32,
     #[serde(default)]
+    pub top_p: Option<f32>,
+    #[serde(default)]
+    pub top_k: Option<u32>,
+    #[serde(default)]
+    pub seed: Option<u64>,
+    #[serde(default)]
+    pub stop: Option<StopSpec>,
+    #[serde(default)]
+    pub frequency_penalty: Option<f32>,
+    #[serde(default)]
+    pub presence_penalty: Option<f32>,
+    /// OpenAI `logprobs` (bool). When true, per-token logprobs are returned
+    /// (engine-permitting).
+    #[serde(default)]
+    pub logprobs: bool,
+    /// OpenAI `top_logprobs` (0..=20): how many alternatives per position.
+    #[serde(default)]
+    pub top_logprobs: Option<u32>,
+    #[serde(default)]
     pub stream: bool,
+    #[serde(default)]
+    pub stream_options: Option<StreamOptions>,
     /// Hybrid-reasoning switch (Qwen3+ convention). Default true =
     /// model-default behavior; false asks the engine to skip the
     /// <think> block (engines that can't, ignore it).
     #[serde(default = "default_true")]
     pub enable_thinking: bool,
+}
+
+impl ChatCompletionRequest {
+    /// Build the engine-facing `SamplingParams`, applying OpenAI defaults
+    /// (top_p=1.0, penalties=0.0) when a field is omitted.
+    fn sampling_params(&self) -> cascadia_types::SamplingParams {
+        cascadia_types::SamplingParams {
+            top_p: self.top_p.unwrap_or(1.0),
+            top_k: self.top_k.unwrap_or(0),
+            seed: self.seed,
+            frequency_penalty: self.frequency_penalty.unwrap_or(0.0),
+            presence_penalty: self.presence_penalty.unwrap_or(0.0),
+            stop: self
+                .stop
+                .as_ref()
+                .map(|s| match s {
+                    StopSpec::Single(x) => vec![x.clone()],
+                    StopSpec::Multiple(v) => v.clone(),
+                })
+                .unwrap_or_default(),
+        }
+    }
+
+    /// Top-logprobs count to request from the engine (0 = disabled).
+    fn logprobs_count(&self) -> u32 {
+        if self.logprobs {
+            self.top_logprobs.unwrap_or(1).clamp(1, 20)
+        } else {
+            0
+        }
+    }
 }
 
 fn default_true() -> bool {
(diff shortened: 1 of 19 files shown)


t8:
Addressed the review findings in 2432d14 (all were LOW/NIT — no blocking bugs):
- **#2/#3:** added `completions_finish_reason_length_when_capped` and `completions_engine_error_returns_503` to cover the forked `length`-finish_reason and 503 engine-error branches (previously inspection-only).
- **#1:** documented that `CompletionChoice.logprobs` reuses the chat-style `ChatLogprobs` shape rather than OpenAI's legacy `{tokens, token_logprobs, text_offset}` object — moot until an engine emits per-token logprobs (none do yet); converting to the legacy schema is a noted follow-up.

The reviewer verified echo race-freedom, stop-truncation UTF-8 safety, permit/InFlightGuard discipline, and SSE validity by inspection — no changes needed there.

Issue #16, opened 2026-05-18
Prometheus /metrics endpoint with request, generation, engine, transport, and MoE metrics


t8:
## Summary

Add a `/metrics` endpoint serving Prometheus exposition format from `tahoma-api`. Today the codebase emits only `tracing` info logs; there's zero hits for `prometheus`, `otlp`, `opentelemetry`, `metric`, `histogram` across `crates/`.

Production observability is missing for: request rate, TTFT, inter-token latency, e2e duration, queue depth, in-flight tasks, model load duration, per-stage transport bytes/latency, cancellation count.

## Why

Prometheus `/metrics` is the de-facto industry standard for inference servers. OVMS exposes it built-in (`docs.openvino.ai/2026/model-server/`). Without it, Tahoma is a black box in any production deployment — operators can't size queues, can't alert on latency regressions, can't compare cost per token across nodes, and can't validate that pipeline-parallel stages are balanced.

This is also a prerequisite for several downstream issues (request cancellation observability, multi-tenant fairness when continuous batching lands, capacity planning for sparse-MoE).

### (1) Dependency

Add to workspace `Cargo.toml`:
```toml
prometheus = "0.13"  # or metrics + metrics-exporter-prometheus
```

Choice between crates:
- **`prometheus = "0.13"`** — battle-tested, used in many production servers (Linkerd, etc.), has its own registry
- **`metrics = "0.23"` + `metrics-exporter-prometheus = "0.15"`** — more idiomatic Rust facade, lets you swap exporters

Recommendation: `prometheus` direct. Simpler, no facade indirection, fewer moving parts. The user can revisit if multiple exporters become a goal.

### (2) Register the route

**File:** `crates/tahoma-api/src/lib.rs` near the route table (`:143-149`)

```rust
.route("/metrics", get(metrics_handler))
```

```rust
async fn metrics_handler() -> impl IntoResponse {
    let metric_families = prometheus::gather();
    let encoder = TextEncoder::new();
    let mut buffer = Vec::new();
    encoder.encode(&metric_families, &mut buffer).unwrap();
    (
        [(header::CONTENT_TYPE, encoder.format_type())],
        buffer,
    )
}
```

### (3) Define metrics

A baseline set to ship in v1:

**Request-level (in `tahoma-api`):**

| Name | Type | Labels | Description |
|---|---|---|---|
| `tahoma_http_requests_total` | counter | `endpoint`, `status` | HTTP requests by endpoint + status code |
[...]

PR #93, opened 2026-06-24
feat(engine): implement engine-side cancel() propagation (#15)

### LINKED ISSUE #15: Implement engine-side cancel() propagation (cancel is currently a no-op)
## Summary

The HTTP API has a `/v1/cancel/:task_id` endpoint (`crates/tahoma-api/src/lib.rs:552-558`) and the runner calls `engine.cancel(task_id)` when an SSE client disconnects (`crates/tahoma-runner/src/lib.rs:327-341`). **But no engine actually implements `cancel()` — it's a no-op default method on the `Engine` trait.** In-flight OV inference completes regardless of client state; only buffered output is dropped.

Implement `cancel(task_id)` in every engine so an in-flight token loop can short-circuit when the client disconnects or explicitly cancels.

## Why

Today a client that opens an SSE stream, gets bored, and disconnects after one token still costs Tahoma the full `max_tokens` worth of compute. On a multi-stage K2.6 pipeline that's tens of seconds per token × multiple boxes of wasted work. On a busy API, this also blocks the next pending task from starting until the abandoned one finishes.

The semantics in OAI's spec: a server SHOULD stop generation as soon as the client closes the connection. Tahoma honors that at the API layer but the engine doesn't get the message.

## Current state

**`crates/tahoma-engine/src/lib.rs:61-81`** — `Engine` trait:
```rust
#[async_trait]
pub trait Engine {
    async fn warmup(&mut self) -> EngineResult<()>;
    async fn submit(&mut self, task: GenerationTask) -> EngineResult<()>;
    async fn step(&mut self) -> EngineResult<Vec<(TaskId, Chunk)>>;
    /// Mark a task for cancellation. Default no-op.
    async fn cancel(&mut self, _task_id: TaskId) -> EngineResult<()> { Ok(()) }
    async fn close(&mut self) -> EngineResult<()>;
}
```

`crates/tahoma-runner/src/lib.rs:327-341` calls `engine.cancel(task_id)` on stream-Drop. `crates/tahoma-api/src/lib.rs:552-558` exposes `/v1/cancel/:task_id`.

None of the four engines override `cancel`.

## Implementation

The key invariant: `cancel(task_id)` must be cheap (sync-ish, idempotent, can be called concurrently with `step()`). It should not block waiting for the task to actually finish — it sets a flag, `step()` observes the flag and exits the inner loop.

### Cancellation surface per engine

| Engine | Cancellation strategy |
|---|---|
| `mock` | Simple flag; `step()` returns the cancellation marker on the next call |
| `ov-genai` | The OV-GenAI `LLMPipeline::generate` runs synchronously inside `step()`. Cancellation requires either (a) the `StreamerBase` callback returning a stop signal — OV-GenAI supports this — or (b) wrapping in `tokio::task::spawn_blocking` and tracking a cancellation token. **Use (a)**: pass a callback that checks a per-task `AtomicBool` and returns `StreamingStatus::CANCEL` |
| `ov-runtime` | Tahoma drives the token loop; check the cancellation flag at the top of each inner iteration in `step()` (`crates/tahoma-engine-openvino/src/runtime.rs:574-575` area) |
| `ov-dist-spec` | Same approach as `ov-runtime`. Additional consideration: when canceling on the driver (rank 0), broadcast a `Reset` to all workers to clear partial state |
| `sparse-moe` | Same as runtime; check flag at top of `Runner::step_distributed` body. For pipeline-parallel: send a `Reset` upstream/downstream to flush partial state |

### Concrete code shape

In each engine struct, add:
```rust
cancelled: HashSet<TaskId>,
```

`cancel(task_id)` inserts:
```rust
async fn cancel(&mut self, task_id: TaskId) -> EngineResult<()> {
    self.cancelled.insert(task_id);
    Ok(())
}
```

`step()` checks at the top:
```rust
if self.cancelled.remove(&current_task.id) {
    return Ok(vec![(current_task.id, Chunk::final_with_reason(FinishReason::Cancelled))]);
}
```

`finish_reason: "cancelled"` is a non-standard OAI value — translate to `null` or `"stop"` at the API layer per spec, and surface as `"cancelled"` only in metrics.

### Pipeline-parallel cancellation

For multi-stage `sparse-moe` / `ov-runtime` / `ov-dist-spec`: rank 0 owns the cancellation decision. On cancel:
1. Set local flag
2. Send `Reset` frame downstream (`crates/tahoma-engine-sparse-moe/src/dist.rs:242` for sparse-MoE, `crates/tahoma-engine-openvino/src/dist_spec.rs` for dist-spec)
3. Downstream ranks reset their KV cache and drain any pending Forward frame
4. Wait briefly for downstream ack (best-effort; on timeout, drop the connection and let reconnect logic re-establish)

Non-rank-0 nodes can be canceled only by rank 0 — they don't have task IDs, just sequenced frames.

## Acceptance criteria

- [ ] Each engine implements `Engine::cancel(task_id)` (no longer the default no-op)
- [ ] After `cancel`, `step()` returns within one inference iteration with a final chunk marked `FinishReason::Cancelled`
- [ ] SSE client disconnect mid-generation causes engine to stop generating within one token (verified by log inspection — no further "decode step" log lines after disconnect)
- [ ] `POST /v1/cancel/:task_id` returns 200 and stops the task within one token
- [ ] For multi-stage `sparse-moe`: cancel on rank 0 propagates a Reset to downstream ranks; verified by 2-box test
- [ ] No regression on the K2.6 3-prompt quality eval (cancellation path doesn't corrupt the normal generation path)
- [ ] Metric: `tahoma_tasks_cancelled_total` counter (depends on #N for prometheus)

## Effort

~ 1 week. Per-engine work is straightforward; the multi-stage propagation needs care.

## Out of scope

- Server-side timeout cancellation (e.g., "cancel any task that's been running > 10 min") — separate small issue
- Resource cleanup of the OV `InferRequest` state machine — `reset_state()` already handles this; cancellation just calls it earlier
- Cooperative cancellation tokens propagated through the entire async call stack — overkill; flag + check is enough

## References

- OAI cancellation semantics: [OpenAI Streaming docs](https://platform.openai.com/docs/api-reference/streaming)
- OV-GenAI `StreamerBase` callback: [openvino_genai header](https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/include/openvino/genai/streamer_base.hpp)
- Frame versioning convention for multi-stage reset: `crates/tahoma-engine-sparse-moe/src/dist.rs:30`

## Labels

`bug`, `enhancement`

t8:
Closes #15.

> **Stacks on #91 (#14)** (needs `FinishReason::Cancelled`). Top commit is the #15 change; merge after #90 → #91.

`Engine::cancel(task_id)` was a default no-op on every engine except qwen36, so a cancelled or disconnected request kept burning compute until its generation finished — on a multi-node K2.6 pipeline that's tens of seconds/token of wasted fleet work. This implements it per engine, following qwen36's proven between-steps idiom (`cancel` and `step` never overlap — both `&mut self` behind the runner mutex).

## Per-engine behavior
| Engine | Shape | cancel() does |
|---|---|---|
| mock, ov-runtime, gemma4, ov-dist-spec (driver) | **step-wise** (`active` + `pending`) | drops the queued task **and** clears the in-flight `active` one → `step()` stops emitting within one iteration |
| ov-genai, sparse-moe (dense + MiniMax) | **monolithic** (one whole generation per `step()`) | drops queued tasks; an already-running generation holds the mutex to its step boundary |
| ov-dist-spec (worker rank) | frame-sequenced, no task IDs | keeps the default no-op |

The runner already tracks cancelled tasks and terminates the client stream; this PR stops the wasted **engine compute** behind it.

## Honest limits (documented in-code as follow-ups)
- **Mid-generation interruption of monolithic engines** isn't solved here: ov-genai needs a `StreamerBase` callback returning `CANCEL` (C++ shim), and sparse-moe single-stage needs an out-of-mutex `Arc<AtomicBool>` cancel token threaded into the decode loop (the engine mutex means `cancel()` can't run *during* a monolithic `step()`). These are the biggest remaining compute-savings wins for long generations.

## Tests
- ✅ mock `cancel_drops_pending_task_and_stops_emission` (emit one token → cancel → `step()` empty; unknown-id cancel is a safe no-op).
- ✅ workspace builds + full test suite green; `cargo fmt --all --check` clean.
- ⏳ The OV-engine cancel paths follow qwen36's already-shipped pattern but aren't exercised by stub tests (need hardware); the AI-PC CI tier compiles them.

### DIFF
--- a/crates/cascadia-api/src/lib.rs
+++ b/crates/cascadia-api/src/lib.rs
@@ -296,6 +296,21 @@ pub struct ChatMessage {
     pub content: String,
 }
 
+/// OpenAI `stop` is either a single string or an array of strings.
+#[derive(Deserialize, Debug)]
+#[serde(untagged)]
+pub enum StopSpec {
+    Single(String),
+    Multiple(Vec<String>),
+}
+
+/// OpenAI `stream_options`. Only `include_usage` is honored today.
+#[derive(Deserialize, Debug, Default)]
+pub struct StreamOptions {
+    #[serde(default)]
+    pub include_usage: bool,
+}
+
 #[derive(Deserialize, Debug)]
 pub struct ChatCompletionRequest {
     pub model: String,
@@ -305,12 +320,64 @@ pub struct ChatCompletionRequest {
     #[serde(default)]
     pub temperature: f32,
     #[serde(default)]
+    pub top_p: Option<f32>,
+    #[serde(default)]
+    pub top_k: Option<u32>,
+    #[serde(default)]
+    pub seed: Option<u64>,
+    #[serde(default)]
+    pub stop: Option<StopSpec>,
+    #[serde(default)]
+    pub frequency_penalty: Option<f32>,
+    #[serde(default)]
+    pub presence_penalty: Option<f32>,
+    /// OpenAI `logprobs` (bool). When true, per-token logprobs are returned
+    /// (engine-permitting).
+    #[serde(default)]
+    pub logprobs: bool,
+    /// OpenAI `top_logprobs` (0..=20): how many alternatives per position.
+    #[serde(default)]
+    pub top_logprobs: Option<u32>,
+    #[serde(default)]
     pub stream: bool,
+    #[serde(default)]
+    pub stream_options: Option<StreamOptions>,
     /// Hybrid-reasoning switch (Qwen3+ convention). Default true =
     /// model-default behavior; false asks the engine to skip the
     /// <think> block (engines that can't, ignore it).
     #[serde(default = "default_true")]
     pub enable_thinking: bool,
+}
+
+impl ChatCompletionRequest {
+    /// Build the engine-facing `SamplingParams`, applying OpenAI defaults
+    /// (top_p=1.0, penalties=0.0) when a field is omitted.
+    fn sampling_params(&self) -> cascadia_types::SamplingParams {
+        cascadia_types::SamplingParams {
+            top_p: self.top_p.unwrap_or(1.0),
+            top_k: self.top_k.unwrap_or(0),
+            seed: self.seed,
+            frequency_penalty: self.frequency_penalty.unwrap_or(0.0),
+            presence_penalty: self.presence_penalty.unwrap_or(0.0),
+            stop: self
+                .stop
+                .as_ref()
+                .map(|s| match s {
+                    StopSpec::Single(x) => vec![x.clone()],
+                    StopSpec::Multiple(v) => v.clone(),
+                })
+                .unwrap_or_default(),
+        }
+    }
+
+    /// Top-logprobs count to request from the engine (0 = disabled).
+    fn logprobs_count(&self) -> u32 {
+        if self.logprobs {
+            self.top_logprobs.unwrap_or(1).clamp(1, 20)
+        } else {
+            0
+        }
+    }
 }
 
 fn default_true() -> bool {
@@ -327,12 +394,21 @@ struct ChatChoiceMessage {
     content: String,
 }
 
+/// OpenAI chat `logprobs` object: one entry per generated token under
+/// `content`. Present only when the request set `logprobs: true` AND the
+/// engine emitted per-token logprobs.
+#[derive(Serialize)]
+struct ChatLogprobs {
+    content: Vec<cascadia_types::TokenLogprobs>,
+}
+
 #[derive(Serialize)]
 struct ChatChoice {
     index: u32,
     message: ChatChoiceMessage,
     finish_reason: &'static str,
-    logprobs: Option<()>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    logprobs: Option<ChatLogprobs>,
 }
 
 #[derive(Serialize)]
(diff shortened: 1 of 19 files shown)


t8:
Reviewed — **no correctness bugs found**, both CI tiers green. The reviewer verified each per-engine claim against the code: field-targeting is correct everywhere, the cancel/step race can't happen (the runner holds the engine lock across the whole `step()` *and* `cancel()`), dropping `active` mid-speculation on dist_spec is safe (next admission's `reset()` wipes the downstream worker's KV while it's quiescent), the monolithic drop-pending is the only possible behavior under the shared mutex, and engine coverage is complete (the dist_spec-worker no-op is correct — it holds no task state).

No code changes needed. One **optional, non-blocking follow-up**: `FinishReason::Cancelled` is currently a reserved-but-unemitted variant — a later change could emit it on the tombstoned final chunk so a cancelled-count surfaces once the Prometheus `/metrics` endpoint (#16) lands. Tracking that with #15/#16 rather than expanding this PR.

Issue #40, opened 2026-05-20
Trained per-layer predictor MLP — PowerInfer-1 §5.1 adaptive sparsity


t8:
## Summary

Train and serve PowerInfer-1-style per-layer predictor MLPs that
forecast which FFN lanes will be active before running gate.
Drop-in replacement for the magnitude-threshold heuristic landed
in PR #34 (CATS/CHESS-style); recovers an additional ~5–10%
sparsity at the same quality budget per PowerInfer §5.1.

## Why

PR #34's two-phase Gate-first sparsity uses a runtime magnitude
threshold (`|silu(gate)| < τ · max_i|silu(gate_i)|`) to decide
which up/down lanes to skip. That's a *post-hoc* signal — we
have to compute gate first to apply it. The PowerInfer-1
predictor (§5.1, Song et al. SOSP'24) trains a *pre-hoc* MLP
that predicts which lanes will be active given just the residual
stream entering the FFN block. The predictor MLP is sized
per-layer based on each layer's sparsity + skewness — typically
6–9% of the total LLM weight.

If the predictor MLP is fast enough (it usually is — ~hidden×1024
+ 1024×intermediate is much smaller than the gate matmul itself
at hidden×intermediate), this saves the per-token cost of running
the full gate matmul on lanes the model would skip anyway.

For K2.6: gate matmul is `intermediate=2048 × hidden=7168 = 14.7M
FMAs` per expert call. A predictor MLP at `hidden × 1024 +
1024 × intermediate = 7M + 2M = 9M FMAs` is 60% the cost — but
since it lets us also skip up/down sparse, the net win can be
>2× over the magnitude-threshold path even without considering
the higher achievable sparsity rate.

### (1) Calibration pipeline

**External tool / Python script** (not part of cascadia main):

- Run K2.6 (or any model) on a calibration corpus (200+ prompts).
- For each layer + expert (or for each shared/dense FFN), record
  the per-token `silu(gate)` outputs.
- Train a per-layer predictor MLP per the PowerInfer-1 schedule:
  start with a fixed `[hidden, h]` first layer, grow `h` until
  predictor accuracy stops improving (~85–90% recall of
  high-magnitude lanes is the standard target).
- Export the trained predictor weights into a sidecar file
  alongside the model.

This is **training-side** work; tooling lives in `cascadia`'s
`tools/` directory or as a separate Python repo. Won't ship in
the main inference path.

### (2) Predictor inference seam

[...]

PR #95, opened 2026-06-29
fix(transport): bounded non-fatal frame-start recv for active token responses

pawanpaudel93:
## What

On the **non-packed** ov-runtime pipeline path:

1. **A bounded, non-fatal frame-start recv for the token reply.** New
   `TransportError::FrameStartTimeout` (deliberately absent from
   `recv_error_is_connection_fatal`), `recv_exact_frame_start_strict`,
   `recv_tensor_token`, `ActivationClient::recv_token(deadline)`.
2. **A per-hop sequence echo.** Each activation is preceded by a *lead* frame
   carrying a monotonic `u32` seq (plus the absolute position on the static
   path); the downstream neighbour echoes that seq back on the token, so a late
   orphan token is discarded instead of being read by the next request.

`recv_hidden_from_upstream` keeps its lenient frame-start wait — that one is
genuinely idle between requests.

## Why (issue #40, reported against a downstream multi-stage deployment)

**The token wait was already deadlined on `main`.** It used
`recv_reply`/`recv_reply_prefill`, whose frame-start runs under `recv_timeout`.
An earlier revision of this PR claimed the wait was "deadline-exempt, bounded
only by the frame-idle ceiling" — that described this branch's original base,
not `main`, and the claim has been removed from the code comments.

What `main` lacks is **survivability**. A reply timeout there surfaces as
`recv_exact timed out`, which both `recv_error_is_connection_fatal` and
`EngineError::is_connection_fatal` classify as fatal, so the transport drops the
socket — and `ActivationClient` dials once, at startup, with nothing anywhere in
this repo to re-dial it. A head that times out on an orphaned token wait (its
chain re-formed under it after a rank restart) therefore loses its downstream
for the life of the process and answers `not connected` to every later request.

So the fix is the **classification**, not the bound: elapsing at the frame start
consumes zero bytes (cancel-safe `read`), leaves the socket frame-aligned, and
returns a retryable error, so the caller releases the per-engine mutex and the
next request can serve. Keeping the socket is what admits a late orphan token —
hence the seq echo, without which that token would be consumed as the *next*
request's and silently desync the stream.

This was measured on a 4-rank chain (see Testing): on `main` the first
post-restart request hangs ~400s and every request after it fails
`not connected`, permanently. On this branch the failure surfaces in 0.1s and
the head stays connected.

## Wire format change (breaking)

The non-packed activation wire changes, and there is no version negotiation —
**all ranks in a pipeline must run the same build**:

| | before | after |
|---|---|---|
| stateful activation | `[hidden]` | `[lead I64 [1,1,1] = seq] [hidden]` |
| static (NPU) activation | `[pos I64 [1,1,1]] [hidden]` | `[lead I64 [1,1,2] = seq, pos] [hidden]` |
| token reply | `I32 [1,1,1]` | `I32 [1,1,2] = [token, seq]` |

The position moved *into* the lead frame deliberately. A standalone seq frame and
a standalone position frame were byte-identical (both `I64 [1,1,1]`, 8 bytes), so
on the static path a pre-upgrade peer's position was silently bound as a sequence
number and the failure surfaced one frame later as a bogus complaint about the
hidden frame. Shape now discriminates, a stage validates the lead against its own
staticness, and skew is a hard error in both directions. It also drops one frame
per token per hop (`set_nodelay` is on, so that was a real extra TCP segment).

Mixed-version skew is rejected with the cause named — verified on hardware:

```
expected an I32 [1,1,2] 8-byte token frame ([token, seq]), got dtype=I32
shape=[1, 1, 1] len=4 - a 4-byte I32 [1,1,1] frame means the downstream runs a
build predating the seq-tagged token wire (upgrade both stages); an I64 frame
means a packed stage is wired to a non-packed one
```

## Review round

A multi-agent review of the rebased branch found five blocking issues and nine
smaller ones. Each is fixed in its own commit:

- **The deadline did not bound the wait.** `recv_tensor_token` charged only the
  frame start; a peer that dribbled a partial header then wedged held the engine
  mutex for `deadline + 2 x recv_timeout` (measured: a 200ms deadline took 3.0s
  at `recv_timeout=3s`; 360s at the rig's 120s config). Every phase now runs
  against one deadline, with a `FRAME_PHASE_GRACE` floor so a first byte landing
  at the edge can't hand the body an epsilon budget and fatally drop the socket.
- **The prefill budget was cut 10x.** `recv_token_from_downstream` ignored its
  `prefill` flag, capping prefill replies at `min(recv_timeout, 120s)` where
  `main` gave `recv_timeout x 10`. Restored, ceiling and budget widening
  together, in a pure `token_recv_deadline` that is actually testable.
- **A relay stage went silent on failure.** `step_last`/`step_middle` returned
  their error without answering an upstream that had already handed over an
  activation group. They now NACK (a negative token on the ordinary seq-tagged
  frame — a distinctly-shaped NACK carries no seq, so a stale one would abort the
  *next*, healthy generation). Middle ranks also escalate to a supervisor rebuild
  after 3 consecutive token-wait timeouts, which the non-fatal classification
  otherwise removed.
- **The token frame was validated by length alone.** A packed stage's single-row
  reply is `I64 [1,1,1]`, 8 bytes — it decoded as `token`=low word, `echo_seq`=0,
  and since the first stamped seq was also 0 the *first* exchange of a
  mismatched pipeline emitted a token. Now dtype + shape + length.
- **The classifier pin pinned nothing.** The test asserting `FrameStartTimeout`
  is non-fatal lived in `cascadia-engine`, which does not depend on
  `cascadia-transport`, and hardcoded its own copy of the message; rewording the
  `#[error]` would have left it green while production flipped to fatal. It now
  builds the string from the real `Display`.

Also: exhaustive `recv_error_is_connection_fatal` (no `_` arm), one seq counter
instead of two, `inbound_seq: Option<u32>`, discard count + last echoed seq in
the terminal error with the per-frame WARN rate-limited, and corrected comments
and wire docs.

## Testing

`cargo fmt` / `clippy` clean; `cargo test --workspace --all-targets`: **594
passed, 0 failed**. `cargo deny check all`: advisories, bans, licenses, sources
all ok. Windows x86_64 `--features openvino` release build: passes.

Rig: 4-rank plain chain on a Lunar Lake AI-PC, Llama-3.2-1B 4-stage, CPU device.

| Scenario | Result |
|---|---|
| Baseline | 4/4 served, coherent, ready 0.5s |
| Tail-restart A/B | `main`: 400s hang, then `not connected` permanently. Branch: 0.1s attributed 503, socket survives |
| Seq guard | Forced a real orphan; discarded with `expected=2 got=1`, 3/3 follow-ups coherent |
| Prefill, compressed | 12.9s prefill served under a 5s base timeout |
| Prefill, default | 27,046-char prompt served in 54.8s |
| Mixed-version | Loud failure in 1.6s, cause and remedy named |

## Known limitations

- **Restart self-heal cannot be validated in this repo.** `Builder::connect` is
  the only production dial site and runs once at startup; nothing re-dials, so a
  restarted rank can never rejoin on *any* build. That scenario depends on
  reconnect machinery that lives outside this repository. What is demonstrable
  here is the precondition: the head's socket survives instead of being dropped.
- After a middle rank is rebuilt, the head's once-dialed downstream client is
  still dead (`ConnectionReset` is fatal, nothing re-dials), so the head needs its
  own restart. Unchanged from `main`.
- The Linux `--features openvino` release build has **not** been verified on this
  branch; the Windows OpenVINO build covers the same code but not the Linux
  toolchain.
- The seq field wiring and the relay escalation counter have no unit coverage —
  they need a construction seam that does not exist or hardware.

### DIFF
--- a/crates/cascadia-engine-openvino/src/runtime.rs
+++ b/crates/cascadia-engine-openvino/src/runtime.rs
@@ -710,10 +710,26 @@ fn encode_token_with_seq(token: i32, seq: u32) -> WireTensor {
 
 /// Decode an I32 `[1,1,2]` token frame into `(token, echo_seq)`. The seq cast
 /// round-trips the bit pattern (i32 -> u32), so wrap-around values survive.
+///
+/// Validation is STRICT — dtype, shape and length, matching `decode_wire_lead`
+/// and `decode_wire_tokens` rather than checking length alone. A length-only
+/// check accepted any 8-byte frame, and the packed path's reply is exactly
+/// that: `encode_wire_tokens` for one row is I64 `[1,1,1]`, 8 bytes. It decoded
+/// as `token` = the low word and `echo_seq` = 0, and since the first stamped
+/// seq was also 0 the FIRST exchange of a non-packed head wired to a packed
+/// neighbour matched and emitted a token from a structurally mismatched
+/// pipeline; every later step then discarded every reply and stalled the full
+/// deadline. A legacy 4-byte I32 `[1,1,1]` token from a pre-seq peer is caught
+/// here too, with the remedy named.
 fn decode_token_with_seq(t: &WireTensor) -> EngineResult<(i32, u32)> {
-    if t.data.len() < 8 {
+    if t.dtype != WireDType::I32 || t.shape != [1, 1, 2] || t.data.len() != 8 {
         return Err(EngineError::Backend(format!(
-            "downstream sent {}-byte token tensor; need at least 8 ([token, seq])",
+            "expected an I32 [1,1,2] 8-byte token frame ([token, seq]), got dtype={:?} \
+             shape={:?} len={} — a 4-byte I32 [1,1,1] frame means the downstream runs a build \
+             predating the seq-tagged token wire (upgrade both stages); an I64 frame means a \
+             packed stage is wired to a non-packed one",
+            t.dtype,
+            t.shape,
             t.data.len()
         )));
     }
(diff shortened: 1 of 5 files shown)


7i7o (approved):
tACK 👍

PR #105, opened 2026-07-10
feat: deepseek v4 support in the sparse-moe engine

pawanpaudel93:
## Summary

Adds end-to-end support for **DeepSeek-V4-Flash** (`deepseek_v4`, ~250B MoE, FP4 experts) to the sparse-moe engine: an exporter, a CPU reference model, the Rust shell, and a distributed pipeline that shards the model one stage per node across 4× 32 GB Intel AI-PCs.

The checkpoint ships a standalone reference implementation (no HF-transformers modeling file), so the shell is implemented directly against DeepSeek's `inference/` reference and golden-tested against it.

## What's included

- **Exporter** (`tools/export_deepseek_v4.py`) — FP4/FP8 dequant, per-expert int4 + shell safetensors, and `--tiny/--med/--big` synthetic modes for the correctness harness; carries the serving sidecars (tokenizer) and, since the checkpoint ships no jinja template, emits an R1-family `chat_template.jinja` so chat completions render correctly.
- **CPU reference** (`tools/deepseek_v4_ref/`) — the upstream reference with its GPU tilelang kernels swapped for pure-torch equivalents, so it runs on CPU as a deterministic ground truth and generates the golden fixtures.
- **Rust shell** (`crates/cascadia-engine-sparse-moe/src/dsv4/`) — everything around the experts (HC-Sinkhorn hyper-connections, MLA-style MQA-on-latent attention, learned KV compression, the DSA indexer, attention sinks, grouped low-rank output). None is OpenVINO-traceable, so it mirrors the CPU reference 1:1. Experts run as mmap int4 (default) or optional per-expert OpenVINO IR.
- **Distributed pipeline** (`src/engine.rs`, `src/dist.rs`) — one stage per node; rank 0 embeds + drives, mid ranks relay the hidden state, the last rank runs the head + sampler.

## Performance

Measured on the 4-node fleet (DeepSeek-V4-Flash, CPU int4 experts): **steady-state decode ~1.06 tok/s** (0.95 s/token) with **~16.6 s TTFT** on a short prompt — a two-length generation (24 vs 78 tokens) separates the fixed prefill cost from per-token decode. Decode is compute-bound (`DSV4_PROFILE`: routed+shared experts ~57%, attention projections ~30%, the rest ~13%); the default (Rust mmap expert) path was optimized ~2×:

- **Fused AVX2 int4 expert kernel** — fuses the int4 nibble decode straight into the dot product (SIMD unpack → FMA, no f32 scratch row). ~1.8×.
- **Parallel grouped `o_proj`** — the previously-serial `wo_a` mid-GEMV is spread across cores (bit-identical). ~+10%.
- **One-way streamed prefill** (`FrameKind::ForwardPrefill`) — pipelines the prompt across ranks instead of a blocking round-trip per token; a mid-prefill failure is made connection-fatal so a dropped frame can't leave a silent KV hole.

Env-gated debug tools (`DSV4_PROFILE`, `DSV4_OV_STATS`) are no-ops when unset.

## Robustness & correctness

The pipeline spans 4 nodes over a relayed link, so one node dropping out must fail cleanly rather than wedge the chain:

- **Bounded pipeline reply-recv** (`dist::recv_token_reply`) — the rank-0 head and every mid relay await the owed token reply on a strict deadline (`recv_timeout`, env/config tunable), not the idle-tolerant frame ceiling. A downstream that dies mid-request — including a black-holed socket with no FIN/RST, the failure mode seen on this fleet — now surfaces as a fast error instead of pinning the stage on the ~900 s idle ceiling. dsv4 forwards one token at a time, so prefill and decode share a single per-token deadline (no batched-prefill widening).
- **Real `/health` readiness** — previously a hardcoded `200`. A completion that fails with a 5xx flips readiness, so `/health` returns `503 {status:"degraded"}` until the pipeline serves a request again — a load balancer / supervisor can observe a broken chain.
- **TCP keepalive on inter-rank sockets** — nodelay + keepalive (30 s idle, 15 s interval) keeps an idle relayed link warm and detects a silently-dropped peer.
- **Clean served model name** — the OpenAI `model` id is the `--model` basename (or `--served-model-name`), not the raw on-disk path, so chat clients address a stable id.
- **Correct `finish_reason`** — a decode cut off by the context window now reports `length` (was `stop`), so a client can tell a truncated answer from a clean one; an over-budget prompt is logged loudly instead of silently dropping its newest tokens.

## Validation

- FP8/FP4 dequant byte-exact vs DeepSeek's `convert.py` + `kernel.py` on a real shard.
- Rust shell matches the CPU reference greedy (`--med/--big` harnesses, real structural dims).
- 2-, 3-, and 4-rank wire chains match the single-process reference; mmap and eager expert dispatch agree.
- Chat template: render + loader goldens (`cascadia-api`) prove the R1 output vs the legacy raw-text fallback and that the export root layout loads; on-fleet `/v1/chat/completions` returns a coherent instruct answer (`What is the capital of France?` → `The capital of France is Paris.`).
- On-fleet inference verified across diverse prompts (facts, arithmetic, code, long prompt).
- **Fault injection (live, 4-node fleet):** killing a mid-rank makes a chat return `503` in ~61 s (bounded by the reply deadline) instead of hanging with no HTTP status; `/health` degrades to `503`; and the watchdog supervisor auto-restarts the chain back to `200` — a full kill → degrade → recover cycle.

## Known limitations / future work

Deliberately out of scope for this bring-up (single model, single request); tracked as follow-ups:

- **No prompt-prefix KV reuse across turns** — each request re-prefills the full conversation (`step_first` resets every rank), so multi-turn chat re-pays TTFT each turn. `KvPrefixCache` exists but is not yet threaded through the distributed path.
- **Recovery is supervisor-driven** — a dead peer tears the stage down and the external watchdog restarts the chain (~2 min); there is no in-process rank-0 reconnect/replay yet.
- **Single in-flight request** — the pipeline serves one request, one token at a time; concurrent requests serialize (no cross-request batching).
- **Decode is attention-bound (~1 tok/s)** — the largest levers are batched-attention prefill and a direct (non-relayed) inter-node link; expert offload to the AI-PC NPU/GPU (OpenVINO backend, opt-in today) is the decode lever.
- **Over-budget prompts truncate the tail** (the newest turn) — logged loudly, but a budget-aware trim or an explicit 4xx would be better.
- **`<think>` reasoning is not split** into an OpenAI-style `reasoning_content` field; it renders inline in `content`.

## Test plan

**Which tier do you need?**
- **Tier A** — verifies *all* the correctness (shell math, quantization, int4 experts, the multi-rank wire protocol, chat-template rendering) with **no model and no hardware**. This is enough for most reviewers.
- **Tier B** — the full 4-node distributed run, for anyone with the hardware.

---

### Tier A · correctness — no model, no hardware

Run from the root of this repo.

**A1 — full gate** (fmt + clippy + every workspace test). Includes the 2/3/4-rank wire protocol over real `cascadia-transport` — the *same* distributed code path, single-process on a tiny model — and the chat-template render/loader tests below:

```bash
just check
```

**A2 — per-layer goldens vs the CPU reference** *(optional; Python + torch, CPU-only)*. Regenerate the gitignored fixture, then the gated tests run:

```bash
python tools/deepseek_v4_ref/gen_fixtures.py \
  --out crates/cascadia-engine-sparse-moe/tests/fixtures/dsv4
```

**A3 — exporter smoke** *(optional; torch, no download)*. FP4/FP8 dequant → int4 repack → manifest on a synthetic model:

```bash
python tools/export_deepseek_v4.py --tiny --out /tmp/dsv4_tiny
```

**A4 — chat-template rendering** *(no model, no hardware; already run by A1)*. Proves `/v1/chat/completions` renders the R1 `<｜User｜>`/`<｜Assistant｜>` prompt vs the degraded legacy fallback, and that the root-layout export (jinja + AddedToken bos/eos) loads:

```bash
cargo test -p cascadia-api chat_template
```

---

### Tier B · full distributed run — 4 nodes

Uses the `cascadia-fleet-tests` harness. **You need:** 4 worker nodes, 1 big-RAM export host, and the gated `deepseek-ai/DeepSeek-V4-Flash` checkpoint.

**Two machines are involved** — each step below is tagged with where to run it:
- **`[host]`** — the big-RAM **export host** (builds the shards).
- **`[orch]`** — the **orchestrator**: your laptop/box with the `cascadia-fleet-tests` checkout + SSH access to the nodes. Run `[orch]` steps from the fleet-tests repo root.

All steps are **idempotent** — safe to re-run (a re-run after verify is a no-op).

#### B0 · configure your nodes  `[orch]`

Node identity lives in **two files** — edit only these:

**(1) `nodes.conf`** — SSH creds + one line per node. **⚠️ Replace `name`, `ip`, and the paths with your own four nodes** (line order = rank 0 → 3):

```bash
SSH_USER=<your-ssh-user>
SSH_KEY="$HOME/.ssh/<your-key>"
NODES=(
  # name | ip | repo_path_on_node | openvino_dir_on_node
  "node1|100.x.x.x|C:\\path\\to\\cascadia|C:\\opt\\openvino"
  "node2|100.x.x.x|C:\\path\\to\\cascadia|C:\\opt\\openvino"
  "node3|100.x.x.x|C:\\path\\to\\cascadia|C:\\opt\\openvino"
  "node4|100.x.x.x|C:\\path\\to\\cascadia|C:\\opt\\openvino"
)
```

**(2) `dsv4/infra.sh`** — the `rank → "node_ip LO HI jump"` `SLICE` map. Point each rank at one of the nodes above (with its layer range), using the **same names/IPs**.

> Build + run dial the `nodes.conf` IPs directly. Only the one-time **offload** routes through the `SLICE` jump host — if your nodes are directly SSH-reachable, that leg needs the direct-copy variant.

#### B1 · export the model  `[host]`

Full download + one-shot export to `$BIG/dsv4-export-full`:

```bash
BIG=/big/partition ./dsv4/dsv4_export_full.sh
```

> Low-disk host? Stream it instead: `python dsv4/export_deepseek_v4_windowed.py --model <ckpt-dir> --out dsv4-export`

#### B2 · distribute + verify  `[orch]`

Copy each rank's slice to its node (`SRC` = the export dir from B1), then verify:

```bash
SRC=/big/partition/dsv4-export-full RANKS="0 1 2 3" ./dsv4/dsv4_offload_all.sh
./dsv4/dsv4_verify_nodes.sh
```

✅ Expect: `ALL RANKS COMPLETE: 4 nodes, 43 layers, every layer has 257 experts`

#### B3 · build on every node  `[orch]`

```bash
./sync-build.sh pawan/deepseek-v4-sparse-moe
```

#### B4 · run + query (chat)  `[orch]`

`node1,…,node4` are your `nodes.conf` names, **in rank order** (rank 0 first). The query goes through `/v1/chat/completions`, which renders the R1 chat template shipped by the exporter:

```bash
GEN_PROMPT='What is the capital of France? Answer in one short sentence.' GEN_MAXTOK=32 \
  ./run-chain.sh dsv4-chat sparse-moe CPU 8180 9280 \
  'C:\cascadia\models\dsv4-4stage' node1,node2,node3,node4
```

✅ Expect a coherent instruct answer, e.g. `The capital of France is Paris.` — a real answer, not a raw-text continuation. (First token is slow: ~16 s TTFT while the prompt streams across the 4-node chain; then ~1 tok/s.)

#### B5 · batteries  `[orch, optional]`

```bash
./dsv4-infer-battery.sh    # diverse prompts + long-prompt streamed prefill
./dsv4-fix-probes.sh       # context bound, seeded-temp reproducibility, empty-prompt safety
```

## Notes

- The checkpoint is **DeepSeek-V4-Flash** (the aligned/instruct model, `DeepseekV4ForCausalLM`). It ships its chat format as a Python encoder (`encoding_dsv4.py`), **not** a jinja `chat_template` — so the exporter emits an R1-family `chat_template.jinja` and the sparse-moe CLI loads it from the model root, and `/v1/chat/completions` renders the correct `<｜User｜>`/`<｜Assistant｜>` prompt. Chat-style messages work (LibreChat, coding agents). Without the template the API falls back to a raw-text join that degrades the instruct model into document-continuation mode.
- Decode is compute-bound (experts ~57%, attention projections ~30%); the inter-node relay dominates **TTFT** (prefill), so the largest latency win is a direct LAN between nodes (removing the relay).

See [`docs/architectures/deepseek-v4.md`](docs/architectures/deepseek-v4.md) for the architecture and design detail.

## Checklist

- [ ] `just check` passes locally (`cargo fmt --check`, `cargo clippy --workspace`, `cargo test --workspace`)
- [ ] Commits follow [Conventional Commits](https://www.conventionalcommits.org/) (`feat:`/`fix:`/`perf:`/`docs:`/`style:`), one logical change per commit
- [ ] Touches an OV engine — tested on real Intel Core Ultra (Lunar Lake) AI-PCs with OpenVINO 2025.x; the Rust mmap expert path is the default and the OpenVINO int4 expert backend is opt-in


7i7o (approved):
tACK 👍

PR #107, opened 2026-07-15
feat: hybrid NPU+CPU execution — chunked prefill on one device, decode on another

t8:
## NPU+CPU hybrid execution: chunked prefill + per-phase device binding on the static path

**Primary use case: faster TTFT on AI-PC silicon, from 1B on one box to 70B-class on a fleet.** This PR ships the mechanism; the measured campaign across model sizes is in `docs/perf/HYBRID_NPU_CPU.md` (device × model-size matrix, big-model routes) and `docs/perf/NPU_TTFT_TIERS.md` (1B→70B tiers, attribution, fleet deployment learnings).

### What ships

- **Chunked-prefill IR variant**: `tools/export_shards.py --static-prefill-seq C` emits a second static graph (seq=C, context aligned so both variants share one host `StaticKv` ring byte-identically). Prompt consumed C tokens per forward: stage weights stream once per chunk instead of once per token.
- **Per-phase device binding**: `cascadia worker --device X --prefill-device Y` — prefill and decode each compile on their own device; the shared host-DRAM KV ring makes the handoff zero-copy. Works per pipeline stage; chunks ride the existing wire as `[1, r≤C, hidden]` frames.
- Over-window parity cap (`chunk_take`), sub-chunking for mixed-C pipelines, graceful degradation for stages without the variant, phase timing in the task-done log, and the parity/bench harness (`tests/static_prefill_parity.rs`) with `CASCADIA_PARITY_SOFT` (near-tie sweeps) and `CASCADIA_STATIC_TASKS` (steady-state TTFT past the NPU cache-import init) knobs.

### Headline measurements (author campaign — Lunar Lake 258V, 32 GB, INT4, steady-state)

| Scale | Config | TTFT short / long | vs tokenwise |
|---|---|---|---|
| 1B | hybrid NPU→CPU | **129 ms / 644 ms** | 12× / 35× |
| 3B | hybrid NPU→CPU | **321 ms / 1.58 s** | 14× / 40× |
| 8B | 2-stage all-NPU pipeline (one box) | 2.0 s / 7.9 s warm e2e | — |
| 14B | chunked GPU (single box) | **603 ms / 2.79 s** | 15× / 41× |
| 32B | 3-box pipeline (iGPU+NPU+NPU) | 6.0 s / 27.7 s warm e2e | — |
| 70B | 6-box pipeline | health ✓; execution blockers isolated & documented | — |

Decode is unchanged in every hybrid config. **Greedy parity is near-tie-tolerant, not token-exact** (under OV 2026.2.x) — see the parity note below. Honest framing of the multipliers (baseline = the seq=1 static path; prior-art attribution incl. Sarathi, NPUW, llm.npu, NITRO) is in NPU_TTFT_TIERS.md.

### Known limitations

Static path only; greedy only; ~2× stage weights resident when phases split across devices (see #108 for residency paths); a tokenwise-prefilling stage can starve pipeline transport past the 60 s frame timeout (documented; keep-alive follow-up); very short prompts (≲C/8) may not gain.

Stacked under #108 → #109. **Do-not-merge hold per repo owner.**

---

## Rebase + re-validation (2026-07-24)

Rebased onto `main` (`c5d4bda`; now includes #105 DeepSeek-V4 + #106 fixes + #111 release). The rebase is **faithful** — `runtime.rs` (the argmax + chunked-prefill inference) is byte-identical pre/post-rebase; the only OV/export-path deltas are a doc comment, an exporter error-message string, and an MSRV line. Independently re-reviewed and re-measured on hardware (matias-04, Lunar Lake 258V, release build).

### Independent A/B (Llama-3.2-1B INT4 static shard, CPU decode, steady-state)

| Prompt | tokenwise (= pre-#107) | chunked (CPU) | **hybrid (NPU prefill + CPU decode)** |
|---|---|---|---|
| short (~55 tok) | 2933 ms | ~350 ms (8.3×) | **~120 ms (≈24×)** |
| long (~230 tok) | 13270 ms | ~1035 ms (12×) | **324 ms (≈41×)** |

Decode tok/s unchanged (~15 short / ~12.6 long). Confirmed both in-harness *and* against a real pre-#107 binary (`main`) serving the same shard (~3202 ms → hybrid ~120 ms). Greedy/argmax decode confirmed in `runtime.rs`.

## Test plan

**Tier A — no Intel hardware (`just check`, Mac/CI):**
- Ring-math equivalence: `chunked_absorb_matches_sequential`, `chunk_take_caps_at_the_kv_window`, `prefill_mask_layout`, `argmax_row_selects_requested_row` — chunked KV state byte-identical to seq=1.
- Load-time validation: `prefill_device_on_stateful_shard_rejected`, `inconsistent_prefill_context_rejected`, `advertised_prefill_variant_missing_file_rejected`, `disable_chunked_prefill_skips_variant`.
- Parity-verdict logic: `parity_verdict_*` (exact / late-near-tie-tolerated / early-fork-hard-fail / soft).
- CLI arg validation (manual): `shard --static-prefill-seq` (requires `--target npu`; 0 or ≥2; ≤ context−1); `worker --prefill-device / --no-chunked-prefill` guards.

**Tier B — Intel AI-PC + `--features openvino`:**
- `cargo test … --test static_prefill_parity` against a `--static-prefill-seq` export: baseline vs chunked vs hybrid, near-tie-tolerant token parity + TTFT/tok/s bench. Env: `CASCADIA_STATIC_SHARDS`, `CASCADIA_PREFILL_DEVICE=NPU`, `CASCADIA_PARITY_SOFT`.

## Parity note: near-tie-tolerant, and OV-version-sensitive

The chunked/hybrid legs run a **different compiled graph** (seq=`C`) than the seq=1 decode graph, so the two accumulate FP differently and can flip a near-equal top-2 argmax — forking the greedy text (both branches coherent, re-converging). The host KV ring state is byte-identical (proven by the ring-math unit tests); the divergence is purely in the compiled graphs' FP, and **which** near-ties flip is **OpenVINO-version-sensitive**.

Measured on the current fleet (**OV 2026.2.0**, greedy/argmax, deterministic), a 1B + 3B sweep over varied prompts found forks **as early as token 2**, even same-device CPU. So the earlier "same-device CPU/NPU output is token-exact (hard-asserted)" statement **does not hold under 2026.2.0** — it may have held on the **2026.1** the measurements above document (not re-verified). Accordingly `assert_parity` now tolerates a near-tie with a loud report and hard-fails only a **token-0** fork (`NEAR_TIE_MIN_PREFIX = 1`, the strongest wrong-KV signal); `CASCADIA_PARITY_SOFT=1` tolerates even that.

This is **not** a rebase or review artifact — verified empirically: the pre-review binary (`2552408`, none of the review changes) forks **identically** to HEAD on the same shard + OV 2026.2.0 (chunked token 2, hybrid token 16, same divergent text). The decode path (`argmax_logits_row` / `chunk_take` / `absorb_layer_multi`) is byte-identical pre/post-rebase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,7 @@ Cascadia is a Cargo workspace; one concern per crate. The `Engine` + `Builder` t
 
 - **Placement is manual today.** Operators set `--rank` / `--total` / `--listen` / `--next host:port` on each worker. `cascadia discover` browses the LAN, but workers still need explicit ranks, full auto-ring formation is not yet wired into `cascadia worker` (tracked in [#89](https://github.com/labscommunity/cascadia/issues/89)).
 - **Device profiling.** `cascadia profile-devices --model <dir>` benchmarks each OV device (iGPU / NPU / CPU) on a host and writes `device_profile.json`, step 1 toward automatic placement. See [docs/perf/DEVICE_PROFILE.md](docs/perf/DEVICE_PROFILE.md).
+- **Hybrid NPU+CPU phase split.** On static (`--target npu`) shards, `cascadia worker --device CPU --prefill-device NPU` runs the compute-bound chunked prefill on the NPU and the bandwidth-bound decode on the CPU, sharing one host KV ring. See [docs/perf/HYBRID_NPU_CPU.md](docs/perf/HYBRID_NPU_CPU.md).
 - **Tensor parallelism:** type-system plumbing only; no engine implements it yet. See [docs/TENSOR_PARALLELISM.md](docs/TENSOR_PARALLELISM.md).
 
 ## Deploying
--- a/crates/cascadia-cli/src/lib.rs
+++ b/crates/cascadia-cli/src/lib.rs
@@ -197,7 +197,7 @@ pub enum Command {
     Completions(CompletionsArgs),
 }
 
-#[derive(ValueEnum, Clone, Copy, Debug)]
+#[derive(ValueEnum, Clone, Copy, Debug, PartialEq, Eq)]
 pub enum EngineKind {
     Mock,
     OvGenai,
@@ -360,6 +360,24 @@ pub struct WorkerArgs {
     /// LLMPipeline).
     #[arg(long, value_name = "TOKS")]
     pub npu_min_response_len: Option<u32>,
+
+    /// Device for the chunked-prefill IR variant (default: same as
+    /// --device). ov-runtime engine on a stateless static export
+    /// (tools/export_shards.py --target npu --static-prefill-seq N) only.
+    /// The phase split: e.g. `--device CPU --prefill-device NPU` runs the
+    /// compute-bound wide prefill on the NPU and the bandwidth-bound seq=1
+    /// decode on the CPU, sharing one host-side KV ring. Accepts the same
+    /// OpenVINO device forms as --device. Note both compilations hold their
+    /// own copy of the stage weights (~2x weight RSS when the devices
+    /// differ); KV is not duplicated.
+    #[arg(long)]
+    pub prefill_device: Option<String>,
+
+    /// Ignore the export's chunked-prefill IR variant and prefill one token
+    /// per step (ov-runtime static path only; conflicts with
+    /// --prefill-device). Escape hatch / A-B baseline knob.
+    #[arg(long, default_value_t = false)]
+    pub no_chunked_prefill: bool,
 
     /// Speculative-decode draft model path (FastDraft companion).
     #[arg(long)]
@@ -576,6 +594,8 @@ impl WorkerArgs {
             npu_prefill_chunk_size: None,
             npu_max_prompt_len: None,
             npu_min_response_len: None,
+            prefill_device: None,
+            no_chunked_prefill: false,
             draft_model: None,
             draft_device: None,
             spec_k: 5,
@@ -636,6 +656,13 @@ pub struct ShardArgs {
     /// static_context - static_seq); ignored without `--target npu`.
     #[arg(long, default_value_t = DEFAULT_STATIC_CONTEXT)]
     pub static_context: u32,
+
+    /// NPU only: also emit a chunked-prefill IR variant with a fixed seq=N
+    /// query window (openvino_prefill_model.xml). Enables `cascadia worker
+    /// --prefill-device` phase-split execution (e.g. prefill on NPU, decode
+    /// on CPU) and ~N× fewer prefill forwards even single-device. 0 = off.
+    #[arg(long, default_value_t = 0)]
+    pub static_prefill_seq: u32,
 
     /// torch default dtype during export. FP16 reduces memory and is
     /// required for `--target npu` (the runtime feeds KV as f16).
@@ -1105,6 +1132,10 @@ fn build_builder(args: &WorkerArgs) -> Result<Box<dyn Builder>> {
         }
         EngineKind::OvRuntime => {
             let mut b = OvRuntimeBuilder::new(&args.model, args.rank, args.total, &args.device);
+            if let Some(dev) = &args.prefill_device {
+                b = b.with_prefill_device(dev);
+            }
+            b = b.with_chunked_prefill_disabled(args.no_chunked_prefill);
             if let Some(dir) = resolve_ov_cache_dir(args.ov_cache_dir.as_deref()) {
                 b = b.with_cache_dir(&dir);
             }
@@ -1341,6 +1372,21 @@ async fn cmd_worker(args: WorkerArgs) -> Result<()> {
             "--rank must be in [0, {}); got {}",
             args.total,
             args.rank
+        ));
+    }
+    // Fail loud, not silent: the phase-split flags are load-bearing when
+    // given, and only the ov-runtime engine implements them.
+    if (args.prefill_device.is_some() || args.no_chunked_prefill)
+        && args.engine != EngineKind::OvRuntime
+    {
+        return Err(anyhow!(
+            "--prefill-device / --no-chunked-prefill require --engine ov-runtime \
+             (the chunked-prefill phase split lives in the static-KV runtime path)"
+        ));
+    }
+    if args.prefill_device.is_some() && args.no_chunked_prefill {
+        return Err(anyhow!(
+            "--prefill-device conflicts with --no-chunked-prefill"
         ));
     }
     let is_first = args.rank == 0;
@@ -1845,6 +1891,24 @@ fn shard_exporter_flags(args: &ShardArgs) -> Result<Vec<String>> {
     if args.target == ShardTarget::Npu && args.default_dtype != ShardDtype::Fp16 {
         return Err(anyhow!("--default-dtype must be fp16 for --target npu"));
     }
+    if args.static_prefill_seq != 0 {
+        if args.target != ShardTarget::Npu {
+            return Err(anyhow!("--static-prefill-seq requires --target npu"));
+        }
+        if args.static_prefill_seq == 1 {
+            return Err(anyhow!(
+                "--static-prefill-seq must be 0 (off) or >= 2 (the chunked-prefill window)"
+            ));
+        }
+        if args.static_prefill_seq > args.static_context.saturating_sub(1) {
+            return Err(anyhow!(
+                "--static-prefill-seq ({}) must be <= --static-context - 1 ({}) — a chunk \
+                 wider than the KV window would evict its own tokens mid-chunk",
+                args.static_prefill_seq,
+                args.static_context.saturating_sub(1)
+            ));
+        }
+    }
     let mut flags = vec![
         "--model".into(),
         args.model.clone(),
@@ -1864,6 +1928,10 @@ fn shard_exporter_flags(args: &ShardArgs) -> Result<Vec<String>> {
         flags.push(args.static_seq.to_string());
         flags.push("--static-context".into());
         flags.push(args.static_context.to_string());
+        if args.static_prefill_seq != 0 {
+            flags.push("--static-prefill-seq".into());
+            flags.push(args.static_prefill_seq.to_string());
+        }
     }
     if let Some(s) = &args.layer_split {
         flags.push("--layer-split".into());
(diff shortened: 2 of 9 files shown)


t8:
## Self-review applied + re-validated on hardware

Ran a max-effort multi-angle review (10 finders → adversarial verify → gap sweep, 15 verified findings) and applied the fixes:

**Correctness**
- **Window-parity cap** (the big one): a single chunk-wide mask can't express per-token sliding-window eviction, so chunk rows past position `static_context−1` attended wider than the tokenwise path — over-window prompts silently diverged. `chunk_take()` now caps chunks at the KV window; the over-window tail steps tokenwise. Chunked output is token-identical to seq=1 in **every** regime.
- `decode_wire_position` rejects negative positions (previously wrapped `position as usize` in the chunk absorb math).
- Mixed `--static-prefill-seq` widths across stages: downstream stages now **sub-chunk** oversized frames instead of erroring at first inference.
- Parity test hard-fails when the export lacks a prefill variant (previously passed vacuously).
- Exporter deletes orphaned `openvino_prefill_model.*` on re-export.

**Efficiency / robustness**
- Head stages skip the whole-`[1,C,vocab]` logits copy+convert for chunks nothing reads: O(prompt×vocab) → O(vocab).
- Chunk inputs convert straight into the reusable buffer (no per-chunk staging allocs); truncate-not-copy on relay outputs.
- `warmup()` exercises **both** compiled models on every stage incl. relays — one-time NPU graph init no longer lands on the first request.
- `build()` cross-checks the variants' primary-input kind; one shared mask writer + port resolver (dedup with equivalence test).

**Re-validation (pawan-01, Lunar Lake 258V, Llama-3.2-1B INT4, all runs token-identical):**

| Leg | tokenwise [CPU] | chunked [CPU] | hybrid [NPU+CPU] |
|---|---|---|---|
| Short (~31 tok) | 1,558 ms | 288 ms | 211 ms (7.4×) |
| Long in-window (~435 tok) | 22,923 ms | 1,629 ms | **687 ms (33.4×)** — 27% better than pre-fix |
| **Over-window (~1200 tok)** | 63,782 ms | 13,521 ms | 11,045 ms (5.8×) — **parity now holds; this leg diverged pre-fix** |

2-stage hybrid pipeline smoke: warm request 0.48 s end-to-end (`prefill_ms=109`), first request 0.69 s (was 0.88 — relay warmup), correct output, clean teardown, memory stable 26.3–26.4 GB free throughout.

Remaining known limitations are documented in `docs/perf/HYBRID_NPU_CPU.md` (placement blind to the phase split; ~2× weight RSS when devices differ — weight-residency work tracked separately).

t8:
## Device × model-size matrix (pawan-01, LNL 258V 32 GB, 2026-07-16)

Ran the full `{prefill NPU} × {decode CPU, GPU, NPU} × {1B, 3B, 8B}` sweep this PR's harness enables (plus chunked/tokenwise baselines per device). Full tables + method notes now live in `docs/perf/HYBRID_NPU_CPU.md` ("Device × model-size matrix"); headline TTFTs in ms, short(~31 tok)/long(~435 tok), **steady-state** (request 2+ of a warm process):

| Config | 1B | 3B | 8B |
|---|---|---|---|
| tokenwise `[CPU]` (old path) | 1,543 / 22,787 | 4,526 / 63,559 | 7,458 / 94,589 |
| chunked `[CPU]` | 240 / 1,654 | 649 / 4,360 | 1,582 / 9,161 |
| **hybrid `[NPU→CPU]`** | **129 / 644** | **321 / 1,582** | — (see below) |
| chunked `[GPU]` | 74 / 391 | 226 / 1,228 | 357 / 1,410 |
| chunked `[NPU]` (all-NPU) | 205 / 1,002* | 393 / 1,965* | — |

Decode tok/s ladder flips with size: 1B NPU 24.5 > GPU 21 > CPU 19 · 3B NPU 9.0 > GPU 8.0 > CPU 6.8 · 8B GPU 6.4 > CPU 4.4 > NPU n/a. (*first-request-after-cache-import numbers.)

Findings baked into the doc/tests by this update:

1. **Steady NPU prefill beats chunked-CPU 1.9–2.8×** (growing with prompt length and model size) at identical decode — hybrid `NPU→CPU` is 11.9–40× over the shipping tokenwise path.
2. **8B is NPU-infeasible on 32 GB**: the NPU compile transient alone exceeded ~23 GB host RAM with nothing else resident (26.9 → 3.6 GB free, watchdog-killed; the hybrid attempt with the CPU model resident hit 0.7 GB). Matches Intel's ">16 GB for >7B" guidance. Transient scales with stage bytes → the viable 8B-NPU route is pipeline-parallel (≈3B-class bytes/stage), untested here. A prefill-first load-order swap was considered and dropped: 3B fits either order, 8B fits neither.
3. **NPU blob-cache import defers ~300–430 ms of driver init to the first inference** — first-request TTFT through `--ov-cache-dir` reads that much high; cold in-process compiles (and request 2+) don't pay it. New `CASCADIA_STATIC_TASKS=N` knob in the parity test measures both. (First theory — PERF_COUNT-poisoned cache blobs — was refuted by a clean-cache reproduction.)
4. **Greedy parity across kernels/devices**: same-device CPU/NPU legs stay token-exact at every size (hard assert unchanged). GPU legs and cross-device hybrid legs can fork at a deterministic argmax near-tie (fork rate grows with model size; one 3B NPU→GPU short fork read distinctly worse — quoted in the doc). New `CASCADIA_PARITY_SOFT=1` reports the fork index + both texts instead of aborting, for perf sweeps only.
5. **GPU is the fastest prefill device at every size** — the mechanism is device-agnostic; `--prefill-device GPU` / all-GPU chunked is the raw-TTFT winner when the iGPU is free. The NPU→CPU split remains the config that leaves both accelerators idle between phases (busy-iGPU / multi-tenant / battery cases).

Fleet note: all runs guarded by a free-RAM watchdog (10 s sampling, 4 GB floor); the two 8B NPU compile attempts were the only kills, and the box recovered to 26.9 GB free both times.

t8:
## Big-model NPU routes — 8B on NPU now works three ways (pawan-01, 32 GB LNL)

Follow-up to the device matrix: the "8B NPU-infeasible" cells are now bounded by three **measured** escape routes (details + sizing rules in `docs/perf/HYBRID_NPU_CPU.md` § "Big-model NPU routes"):

1. **2-stage pipeline-parallel, on-box** — per-stage compiles (2.07 GB INT4/stage) fit when serialized via the new `compile_warm_probe` test (~490 s each, floor 7.7 GB free), then the pipeline starts all ranks concurrently as pure blob-cache imports. Warm e2e (24 new tokens): **all-NPU 2.0 s short / 7.9 s long**; hybrid NPU→CPU 2.3 s / 8.9 s. Steady residency 12/16 GB. Bring-up gotcha fixed on the way: never serialize worker *startup* (a non-first rank blocks in transport accept before engine load — deadlock); warm the cache instead.
2. **AOT cross-compile + blob import** — the NPU compiler is a userspace library that runs with **no NPU present**: `libnpu_driver_compiler.so` from the `intel-driver-compiler-npu` deb, dropped into the OV package as `libopenvino_intel_npu_compiler.so`, plus `NPU_PLATFORM=4000`. Compiled both 8B variants on the Linux miner (23.8–37.9 GB peak *there*), shipped the ~5.8 GB blobs (1.4× INT4), imported on the Windows box via the new `Runtime::import_blob` FFI: **6.9–7.5 s import, free RAM unmoved, zero compile transient on the 32 GB box.** Cross-OS/driver handshake passed with OV pinned 2026.1 both sides. (CACHE_DIR entries are NOT portable — hash includes abs path+mtime; cache loads also miss without a device. Use export_model/import_blob.)
3. **NPUW function folding, on-box monolithic** — `NPU_USE_NPUW=YES,NPUW_FOLD=YES,NPUW_FUNCALL_FOR_ALL=YES,NPUW_ONLINE_PIPELINE=REP,NPUW_DCOFF_TYPE=f16,NPUW_DCOFF_SCALE=YES,NPUW_WEIGHTS_BANK=bank0,NPUW_HOST_GATHER=YES` compiles ONE decoder body instead of 32: the monolithic 8B compiled on-box at ~21.5 GB peak (floor 5.4 GB free), **token-exact parity**, chunked NPU prefill TTFT **1.71 s**. Decode through DCOFF/f16 is only ~1.2 tok/s → folding is a **prefill** engine; decode belongs on CPU/GPU (per-device props plumbing is the follow-up; `NPUW_DQ` INT4-resident is the other lever, unmeasured).

Measured constants for capacity planning: on-box NPU compile transient ≈ **6–9× stage INT4 bytes**; blob import ≈ **1.4×** (and fast). Research-validated but unbuilt: hetero-split prefill (FFN-only on NPU — only one block ever compiles; HeteroLLM 8B prefill 247.9 tok/s on UMA), prefix-KV seeding of the host ring (2.2–3.3× TTFT, no compile), speculative prefill (≤7.66×), weightless blobs (leaner import).

New on the branch since the matrix comment: `compile_warm_probe.rs`, `blob_import_probe.rs` + `cascadia_runtime_import_blob` FFI, and the doc section. All runs watchdog-guarded; box recovered fully every time.

t8:
Autonomous review pass across the stack (#107/#108/#109): 10 finder angles per the repo review process, 28 candidates, 15 findings survived verification — 13 fixed, 2 documented as known limitations (relay-stage park triggers on 1-token workloads; no reload-failure latch). Highlights fixed: per-device prefill plugin properties (hybrid + decode-tuning flags no longer conflict), cross-stage static_context homogeneity validation, park-on-cancel/failure, ring-mutation-before-parked-check ordering, gemv-offload CACHE_DIR strip via any property source, stale-AOT-blob guard, Linux GCC/Clang build (SIMD flags + tbb link name), matcher/fusion shape guards. Full findings ledger in the review report; fixes are the three fix(review) commits on each branch. E2E verified on hardware post-fix: 1B hybrid+parking two-task run under HARD parity (steady hybrid TTFT 118 ms), gemv-offload double-run with cache present (regression test for the cache-poisoning finding — parity holds on run 2), 8B AOT blob import through the new staleness gate (7.5 s), and the 8B 2-stage all-NPU pipeline (warm 1.9 s short / 8.1 s long, correct output). All 13 crate test suites green.

7i7o (approved):
tACK 👍

PR #108, opened 2026-07-15
feat: weight-residency paths for the hybrid split — --park-prefill, NPUW bank probe, in-place GEMV RFC

t8:
## Prefill weight residency: parking, NPUW bank probe, and the in-place-GEMV RFC

Stacked on #107. The phase split holds two compiled models; this PR is the three-path answer to that residency cost, in the order they pay off:

1. **`--park-prefill` (shipped, measured)**: drop the prefill CompiledModel after each prefill; reload from the OV blob cache per request. ~1× steady weights for +~0.7 s/request with `--ov-cache-dir` (292 s cold without — load-time warning added). Parking leg in the parity harness (`CASCADIA_PARK=1`).
2. **NPUW weights-bank probe** (`experiments/npuw-bank-probe/`): shared-bank NPUW compiles of our static shards work; memory verdict documented (DCOFF expands INT4→f16 host-side; process-WS blind to L0).
3. **`docs/rfcs/openvino-inplace-int4-gemv.md`**: the upstream ask, narrowed by the spike's evidence (#109) — mmap-backed execution of grouped-INT4 decompress weights as a documented contract + eliminating the compile-time private peak. Upstream-gap evidence included (upstream oneDNN has no optimized s4 grouped-decompression kernel; OV's fast path is fork-only). Not yet filed.

Context from the 1B→70B benchmark campaign (docs on #107/#109): residency rules measured per device — NPU-resident models cost ≈1.4× INT4 in L0 (the priciest place to keep models), iGPU ≈1×, and the dual-blob pattern is what capped 14B pairs and 70B stage-0 on 32 GB boxes — which is exactly the regime parking (and a pipeline-stage park variant, future work) addresses.

Also here: `CASCADIA_STATIC_TASKS` steady-state knob (first-request cache-import init ≈300–430 ms is measurement noise a long-lived worker never re-pays).

**Do-not-merge hold per repo owner.**

---

## Rebase + validation (2026-07-24)

Rebased onto `main` with #107/#109; faithful (engine inference byte-identical pre/post-rebase). `--park-prefill` validated on hardware (matias-04, Llama-3.2-1B static shard): parking drops the prefill CompiledModel after each request and reloads on the next — `parked#1` 1468 ms → `parked#2` 2056 ms, i.e. the reload adds ~1.4 s to that request (the residency↔latency trade), decode unchanged.

## Test plan

**Tier A — `just check` (Mac/CI):**
- Builder rejects park on a stateful shard / without a chunked-prefill variant: `park_prefill_on_stateful_shard_rejected`.
- `--park-prefill` CLI guards (manual): requires `--engine ov-runtime`; conflicts with `--no-chunked-prefill`.
- RFC (`docs/rfcs/openvino-inplace-int4-gemv.md`) + NPUW-probe notes present.

**Tier B — `--features openvino`:**
- Parity harness `CASCADIA_PARK=1` leg — 2 sequential tasks through one engine, near-tie-tolerant parity, the reload cost visible in task-2 TTFT (set `CASCADIA_OV_CACHE` or the reload is a multi-minute cold compile).
- NPUW weights-bank probe (`npuw_bank_probe.rs`, `CASCADIA_NPUW_PROBE=1`).

## Caveat

The **park-without-cache load-time warning** only fires with `--ov-cache-dir ""` (explicitly disabling the auto-default cache) and has **no unit test** — it's observable only at runtime on a real export.

### DIFF
--- a/crates/cascadia-cli/src/lib.rs
+++ b/crates/cascadia-cli/src/lib.rs
@@ -379,6 +379,16 @@ pub struct WorkerArgs {
     #[arg(long, default_value_t = false)]
     pub no_chunked_prefill: bool,
 
+    /// Park the chunked-prefill model between prefills (ov-runtime static
+    /// path only): after each task's prefill phase the prefill CompiledModel
+    /// is dropped — freeing its resident stage-weight copy, the structural
+    /// memory cost of the two-model split — and re-created from the compile
+    /// cache at the next prefill. Trades a per-task reload (logged as
+    /// reload_ms) for ~1x steady-state weight residency; for memory-tight
+    /// stages. Composes with --prefill-device.
+    #[arg(long, default_value_t = false)]
+    pub park_prefill: bool,
+
     /// Speculative-decode draft model path (FastDraft companion).
     #[arg(long)]
     pub draft_model: Option<String>,
@@ -596,6 +606,7 @@ impl WorkerArgs {
             npu_min_response_len: None,
             prefill_device: None,
             no_chunked_prefill: false,
+            park_prefill: false,
             draft_model: None,
             draft_device: None,
             spec_k: 5,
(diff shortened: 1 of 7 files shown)


7i7o (approved):
tACK 👍

PR #109, opened 2026-07-15
spike: CascadiaInt4Gemv extension op — decode GEMV from the IR mmap (route 1)

t8:
## GEMV-offload spike + AOT blob deployment: the big-model NPU enablers

Stacked on #108. Two related bodies of work:

### 1. AOT blob machinery (the 14B–70B enabler; graduated well beyond "spike")

- **`Runtime::import_blob` FFI + engine consumption**: a `.blob` sibling of a stage IR (from `ov::CompiledModel::export_model`) imports instead of compiling when the device is NPU. The NPU compiler's measured **6–9× INT4-bytes host-RAM transient never runs on the box** — blobs import in ~7 s at ≈1.4× INT4 resident. Cross-compilation runs on any big-RAM Linux host via the standalone `intel-driver-compiler-npu` library + `NPU_PLATFORM` (no NPU required; proven Linux→Windows with OV pinned 2026.1).
- **`tests/compile_warm_probe.rs`**: sequential cache-warming so multi-model/multi-stage bring-up never overlaps compile transients (how 8B 2-stage compiles on one 32 GB box).
- **`tests/blob_import_probe.rs`** + fleet tooling (`experiments/fleet-bench/`): chunked sha-verified transfers, WMI-detached worker starts, file-based benching — the deployment recipe that ran 32B across 3 boxes and validated 70B to six-box health (`docs/perf/NPU_TTFT_TIERS.md`).
- Caveats documented: blobs are driver-generation-sensitive per graph (one 32B stage reproducibly killed a healthy device under VCL 1.32.1 — validate each blob with a real inference at deploy); CACHE_DIR entries are not portable across machines.

### 2. CascadiaInt4Gemv spike (measured to its ceiling, kept as probe + evidence)

OV-extension custom op executing decode INT4 GEMVs from the mmapped IR: 113/113 rewrites, **near-tie-tolerant** token parity, ~69–79% of stock throughput. PERF_COUNT FFI attributed the gap fully (kernel weight-streaming only; stock brgemm ≈51 GB/s is fork-only per the oneDNN embedding probe) → the RFC in #108 is the load-bearing path. Three extension-op traps documented in `experiments/gemv-offload-spike/NOTES.md` (precision-following outputs, CSE weights-tag, no shared scratch under work-stealing), plus the NPU cache-import init warning.

**Do-not-merge hold per repo owner.**

---

## Rebase + review + validation (2026-07-24)

Rebased onto `main` with #107/#108; **faithful** (`runtime.rs` inference byte-identical pre/post-rebase). Full `pr-review-toolkit` sweep (correctness / silent-failure / tests / types / comments) — no ship-blockers; findings fixed. Hardware-validated on matias-04:
- **CascadiaInt4Gemv**: 113 mmap INT4 matmuls execute (load-log confirmed), near-tie-tolerant token parity, **~69% of stock decode throughput** (release) — consistent with "below stock brgemm, kept as probe; the RFC is the load-bearing path". Added `NODE_VALIDATION_CHECK(!weights_tag.empty())` (the CSE-merge → wrong-output guard) after review.
- **Transport send-pacing**: `parse_send_burst` now warns on an unparseable value AND on a below-64 KiB clamp-up (both were silent).
- The OV/FFI path compiles + initializes on the target NPU (`doctor` enumerates CPU/GPU/NPU).

## Test plan

**Tier A — `just check` (Mac/CI):**
- `npu_aot_blob` blob-vs-compile decision + `npu_aot_blob_rejects_blob_older_than_ir`; `import_plugin_strips_cache_dir_keeps_rest`; `--gemv-offload` gating (`gemv_offload_on_stateful_shard_rejected`, `..._on_non_cpu_device_rejected`); transport `parse_send_burst_clamp_table`.

**Tier B — `--features openvino`:**
- gemv-offload cross-kernel parity: `CASCADIA_GEMV_OFFLOAD=1` (confirm `offloaded>0` in the load log, else the parity check is vacuous).
- `gemv_residency_probe.rs`, `compile_warm_probe.rs`, `npuw_bank_probe.rs`.

## Caveats / not-yet-validated

- **AOT `.blob` import round-trip (`import_blob` / `blob_import_probe`) was NOT hardware-validated this pass** — it needs a cross-compiled `.blob`. The FFI compiles and is reviewed; the import handshake on a real NPU is unrun here.
- The CascadiaInt4Gemv matcher/rewrite has no standalone test (the only signal it fired is the `offloaded` count); numeric parity vs oneDNN is empirical only, and its throughput is fork-only per the embedding probe.
- **`build.rs` compiles the shim TU at the AVX2/FMA/VNNI baseline**, so `gemv_offload.cpp`'s runtime `cpu_has_avx2_fma()` scalar fallback is dead on a non-AVX2 x86_64 CPU. Harmless under the current **Intel-AI-PC-only** deployment scope (all targets have AVX2+VNNI); noted as dead-code to remove if scope broadens.

### Follow-ups (tracked, not in this PR)
Stale-blob **content-hash provenance** — the mtime guard is defeated by the `copy /b` deploy path, so a stale-weights blob can still import (import a blob → validate with a real inference at deploy); non-UTF-8 stage path → misleading "cannot open" error; a stub unit test for the transport paced-send loop.

### DIFF
--- a/crates/cascadia-cli/src/lib.rs
+++ b/crates/cascadia-cli/src/lib.rs
@@ -1390,14 +1390,10 @@ fn install_shutdown_signal_handler(
     })
 }
 
-async fn cmd_worker(args: WorkerArgs) -> Result<()> {
-    if args.rank >= args.total {
-        return Err(anyhow!(
-            "--rank must be in [0, {}); got {}",
-            args.total,
-            args.rank
-        ));
-    }
+/// Reject phase-split flag combinations before a worker binds any socket. Pure
+/// function of the args so it is unit-testable (the guards were inline in the
+/// async `cmd_worker`, reachable only by running a full worker).
+fn validate_worker_runtime_flags(args: &WorkerArgs) -> Result<()> {
     // Fail loud, not silent: the phase-split flags are load-bearing when
     // given, and only the ov-runtime engine implements them.
     if (args.prefill_device.is_some()
@@ -1417,6 +1413,18 @@ async fn cmd_worker(args: WorkerArgs) -> Result<()> {
             "--prefill-device / --park-prefill conflict with --no-chunked-prefill"
         ));
     }
+    Ok(())
+}
+
+async fn cmd_worker(args: WorkerArgs) -> Result<()> {
+    if args.rank >= args.total {
+        return Err(anyhow!(
+            "--rank must be in [0, {}); got {}",
+            args.total,
+            args.rank
+        ));
+    }
+    validate_worker_runtime_flags(&args)?;
     let is_first = args.rank == 0;
     let is_last = args.rank == args.total - 1;
 
@@ -2197,6 +2205,36 @@ mod python_tests {
         let mut a = WorkerArgs::single_node(model.into(), "GPU".into(), engine, ":8000".into());
         a.engine = engine;
         a
+    }
+
+    /// The phase-split flags only exist in the ov-runtime static path; using
+    /// one on another engine is rejected loudly (here: --gemv-offload on
+    /// ov-genai), not silently ignored.
+    #[test]
+    fn worker_flags_reject_phase_split_without_ov_runtime() {
+        let mut a = worker("m", EngineKind::OvGenai);
+        a.gemv_offload = true;
+        let err = validate_worker_runtime_flags(&a).unwrap_err().to_string();
+        assert!(err.contains("ov-runtime"), "{err}");
+    }
+
+    /// A prefill device (or parking) is meaningless with chunked prefill
+    /// disabled — the two are mutually exclusive.
+    #[test]
+    fn worker_flags_reject_prefill_device_conflict_with_no_chunked() {
+        let mut a = worker("m", EngineKind::OvRuntime);
+        a.prefill_device = Some("NPU".into());
+        a.no_chunked_prefill = true;
+        let err = validate_worker_runtime_flags(&a).unwrap_err().to_string();
+        assert!(err.contains("conflict"), "{err}");
+    }
+
+    /// A valid phase split (ov-runtime + prefill device, chunked enabled) passes.
+    #[test]
+    fn worker_flags_accept_valid_phase_split() {
+        let mut a = worker("m", EngineKind::OvRuntime);
+        a.prefill_device = Some("NPU".into());
+        assert!(validate_worker_runtime_flags(&a).is_ok());
     }
 
     #[test]
--- a/crates/cascadia-cli/src/lib.rs
+++ b/crates/cascadia-cli/src/lib.rs
@@ -2550,6 +2550,96 @@ mod tests {
         assert!(!flags.iter().any(|f| f == "--static-context"));
     }
 
+    /// `--static-prefill-seq` needs `--target npu` (chunked prefill is a
+    /// static-KV-path feature); the default cpu-gpu target rejects it.
+    #[test]
+    fn shard_flags_static_prefill_seq_requires_npu() {
+        let args = parse_shard(&[
+            "cascadia",
+            "shard",
+            "--model",
+            "m",
+            "--output-dir",
+            "o",
+            "--num-stages",
+            "1",
+            "--static-prefill-seq",
+            "4",
+        ]);
+        let err = shard_exporter_flags(&args).unwrap_err().to_string();
+        assert!(err.contains("--target npu"), "{err}");
+    }
+
+    /// `--static-prefill-seq 1` is nonsense (a 1-wide chunk is the seq=1 decode
+    /// path); only 0 (off) or >= 2 is valid.
+    #[test]
+    fn shard_flags_static_prefill_seq_one_rejected() {
+        let args = parse_shard(&[
+            "cascadia",
+            "shard",
+            "--model",
+            "m",
+            "--output-dir",
+            "o",
+            "--num-stages",
+            "1",
+            "--target",
+            "npu",
+            "--static-prefill-seq",
+            "1",
+        ]);
+        let err = shard_exporter_flags(&args).unwrap_err().to_string();
+        assert!(err.contains(">= 2"), "{err}");
+    }
+
+    /// A chunk wider than the KV window would evict its own tokens mid-chunk:
+    /// `--static-prefill-seq` must be <= `--static-context - 1`.
+    #[test]
+    fn shard_flags_static_prefill_seq_exceeds_context_rejected() {
+        let args = parse_shard(&[
+            "cascadia",
+            "shard",
+            "--model",
+            "m",
+            "--output-dir",
+            "o",
+            "--num-stages",
+            "1",
+            "--target",
+            "npu",
+            "--static-context",
+            "8",
+            "--static-prefill-seq",
+            "8",
+        ]);
+        let err = shard_exporter_flags(&args).unwrap_err().to_string();
+        assert!(err.contains("evict"), "{err}");
+    }
+
+    /// A valid `--static-prefill-seq` is forwarded to the exporter verbatim
+    /// (a dropped/defaulted value would silently export a wrong-shape variant).
+    #[test]
+    fn shard_flags_static_prefill_seq_forwarded() {
+        let args = parse_shard(&[
+            "cascadia",
+            "shard",
+            "--model",
+            "m",
+            "--output-dir",
+            "o",
+            "--num-stages",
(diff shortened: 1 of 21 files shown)


t8:
## Throughput push: 66–73% → **75–79% of stock**, parity held at every step

| Config | decode tok/s | vs stock 18.9 |
|---|---|---|
| spike v1 (single-acc AVX2) | 12.4–13.9 | 66–73% |
| **4-accumulator AVX2 + blocked parallel_for (new default)** | **14.2** | 75% |
| **+ AVX-VNNI dyn-quant int8 acts (`CASCADIA_GEMV_VNNI=1`, opt-in)** | **15.0 (14.9 hybrid)** | 79% |

Hybrid TTFT improved to **186 ms** (best measured). Greedy token-parity vs the stock kernel re-asserted after every change — including under VNNI dynamic quantization (kept opt-in anyway: numerics differ by construction).

**Rejected with data** (documented in NOTES.md): a private TBB-independent worker pool (wake latency + oversubscription — 11.8 tok/s; also fixed a straggler use-after-free before measuring) and `INFERENCE_NUM_THREADS=8` (E-cores gate every fork-join lane; hurts stock too — OV's P-core-only LATENCY default is right on Lunar Lake).

**Frontier mapped:** single-core kernel 5.7 GB/s × 3.7 scaling on the 4-P-core arena ≈ 22 GB/s; VNNI barely moved the aggregate → per-node fork-join + memory-level parallelism now dominate, not compute. Floor analysis: kernel 27.9 ms/token + ~39 ms non-kernel vs stock's 52.9 total ⇒ parity needs ~44 GB/s kernel or fewer/larger nodes. The identified next lever is **graph-level fusion in the pass** (one op for q/k/v, one for gate/up → 113 → ~66 nodes, doubles mean GEMV size) — scoped as a follow-up rather than stretched into this spike.

t8:
## Fusion follow-up: built, parity-clean — and measured **perf-neutral**, which changes the roadmap

The op now supports multi-segment weights (one op for a layer's q/k/v, one for gate/up) with a post-rewrite stage that groups siblings sharing an activation and splices a `VariadicSplit` back to the original consumers. Structurally it did exactly what was designed: **113 → 65 ops/token**, token-parity held (`CASCADIA_GEMV_NOFUSE=1` to A/B).

**Result: throughput flat** — fused 14.23 vs 14.16 tok/s; fused+VNNI 14.84/14.72 vs 15.0/14.9; kernel efficiency unchanged at ~21–22 GB/s. The per-node fork-join hypothesis from the previous frontier analysis is **refuted by measurement**.

Revised frontier (NOTES.md updated):
1. The kernel wall (~29 ms/token) is a **~22 GB/s weight-streaming ceiling on the 4 P-cores** — insensitive to compute (VNNI flat) and node count (fusion flat). Stock's blocked repack simply streams with better prefetch locality. Next levers: software prefetch / NT loads / per-thread weight striping — or conceding hot loops to a repacked cache, which reintroduces the residency this spike exists to avoid.
2. The ~27 ms/token graph-side delta vs stock needs **per-node attribution** (OV `PERF_COUNT` through new shim FFI) before more optimization-by-hypothesis.

Fusion stays default-on (fewer nodes, no cost, base for future epilogue fusion). Spike stands at **75% default / 79% VNNI of stock decode**, ~0 resident decode-weight cost, TTFT 186–190 ms hybrid, parity everywhere. Recommend gating further perf work on the PERF_COUNT data.

t8:
## PERF_COUNT instrumentation landed — the remaining gap is now fully attributed, and two prior beliefs died honest deaths

New (kept, production-usable): `Runtime::profiling()` FFI over `get_profiling_info()` + `CASCADIA_PERF_DUMP=1` aggregated per-node dump (enable with `PERF_COUNT=YES` through the property passthrough).

**Attribution, stock vs offloaded decode infer:**

| node kind | stock | offloaded |
|---|---|---|
| weight matmuls | `FullyConnectedCompressed` ×113: **11.2–12.9 ms** (~51 GB/s brgemm) | `Reference` ×65 (our ops): **31.9 ms** (~22 GB/s) |
| attention subgraphs | 13.5–16 ms | 13.9 ms — identical |
| Broadcast/Concat/RoPE/etc | ~5.7 ms | ~5.8 ms — identical |

Corrections the data forced: (1) stock matmuls are FAST — my ~45 ms estimate was wrong; (2) the earlier "~27 ms graph overhead" dissolves — ~25 ms/token of KV-ring copies + engine glue is **shared by both paths**, and the custom ops add zero Converts/Reorders. The offload's entire deficit is kernel streaming: 31.9 vs 11.2 ms.

**Kernel levers this round** (parity-clean, measured ~flat): flat no-helper loops + forceinline (+7% single-core only), software prefetch, 2-row register blocking, VNNI de-branching. Combined with earlier rounds, every cheap lever now has a number: single-core sits at ~6 GB/s vs oneDNN's ~12.8/core for the same packed-INT4 stream.

**Endgames (NOTES.md):** (1) embed a dnnl INT4-weights-decompression matmul primitive over OUR mmapped memory — oneDNN streaming with residency intact; a dependency/build project, the recommended path if the last ~4 tok/s matters; (2) the upstream RFC. Bonus finding that outranks both for the whole pipeline: attention subgraphs (13–16 ms) and the shared ~25 ms/token ring-copy overhead are the static path's real costs — they apply to stock too and are a separate, larger workstream.

Standing results: **77% of stock default / 79% VNNI, ~0 resident decode-weight cost, hybrid TTFT 188 ms, token-parity everywhere.**

t8:
## oneDNN-embedding endgame: built, and **closed by data** — the fast kernels are fork-only

Full dnnl-inside-the-op integration landed (feature-gated: `DNNL_DIR` at build, `CASCADIA_GEMV_DNNL=1|2` at runtime, permanent fallback on failure): matmul primitives with s4 weights + grouped f16 decompression scales built **over the op's mmapped weight bytes** — zero weight copy (our packed `[N,K]` rows are dnnl's plain `ba` for `[K,N]`), scales transposed once to a resident `[G,N]` buffer (~4 MB), `fpmath(f16, apply_to_int)`, TBB-runtime dnnl 3.12 sharing the SDK's `tbb12` (no OMP oversubscription).

**Results:**
- Numerically **correct end-to-end** — token-parity through full generations (layout, nibble order, scale semantics all validated).
- Mode 1 (our layout, zero-copy): `ref`-slow, ~165 ms/op.
- **Mode 2 (`format_tag::any` + one resident reorder — dnnl's own preferred layout): STILL `ref:any`.**

That last line is the finding: **upstream oneDNN has no optimized kernel for s4 grouped-decompression matmul on this shape at all** — at any layout, at any residency cost. OpenVINO's ~51 GB/s `FullyConnectedCompressed` path lives in the openvinotoolkit/oneDNN **fork** (or OV-internal JIT) only. Embedding upstream dnnl cannot reach plugin throughput, period.

Consequences (NOTES.md + RFC updated):
- The RFC is now the load-bearing path and gained sharp evidence: the capability exists, is Intel-maintained, and is absent upstream — exposing or upstreaming it is the only way third parties reach it without forking.
- Alternatives ranked and rejected for now: building OV's oneDNN fork against unstable internal APIs; a multi-week hand-written brgemm-class INT4 kernel.
- The dnnl integration stays in-tree as the reproducible probe (zero cost without `DNNL_DIR`).

**Spike final standing: 77% of stock decode default / 79% with VNNI, ~0 resident decode-weight cost, hybrid TTFT 186–190 ms, token-parity everywhere — and a fully-measured map of every path to the remaining 21%.**

t8:
70B follow-up (autonomous debugging session): both memory blockers fixed (stage-0 dual-blob L0 OOM; tokenwise-fallback transport starvation — timeout env-tunable). Remaining blocker isolated with hop-by-hop debug traces: ~750 KB inter-stage hidden frames intermittently black-hole on DERP-relayed tailscale links (position frame arrives, hidden frame doesn't; varying link per run). Ruled out: graphs (0.2 s infers in isolation), all rank-0 device configs, disk pressure (pawan-01 had hit 0 bytes free — cleaned), tailscale session staleness, send pacing. 32B (~470 KB frames) re-validated clean same-day on the same links. Full dossier + repro assets in docs/perf/NPU_TTFT_TIERS.md and experiments/fleet-bench/. Debug traces + CASCADIA_SEND_BURST_BYTES knob kept on the branch.

7i7o (approved):
tACK 👍

PR #110, opened 2026-07-20
feat(#34): cross-chain KV coordination plane

pawanpaudel93:
## Summary

Runtime / engine half of issue #34 Option C — the cross-chain KV warm-resume plane. Provides the wire contract and the engine-side KV surfaces that the enterprise plane pulls over.

Pairs with **labscommunity/cascadia-enterprise#53** (the enterprise half), which must pin this branch's tip (`b320fac6`). They ship together; #53 re-pins to the merged rev once this lands.

Scope vs `main`: 76 files, ~+15.2k / −263, across 125 commits.

## What's here

- **`cascadia-kv-wire`** (new crate) — the opaque, self-describing KV blob wire contract: `Manifest` + per-layer K/V pairs, codec validate/roundtrip, golden conformance vectors. Shared with the enterprise plane.
- **Engine `KvCoordination`** on the sharded OV engines (ov-runtime, ov-dist-spec, qwen36, gemma4, sparse-moe) — `lookup`/`export` (NEGOTIATE/GET), `insert`, `stash_downstream_rank`, `apply_warm_resume`.
- **Lock-free KV holder** — `OvKvHolder` over a `SharedKvCache` that every capture mirrors into, so a **busy** source engine answers cross-chain pulls without taking the inference lock (this is what makes pull-HIT-under-load work).
- **Plane warm-resume (Phase A/B)** — `WarmResumeTrigger`/`Confirm` wire + `step_first` plane-restore mode (skips the chain RESTORE) + multi-stage per-rank KV distribution.
- **Model-level fingerprint** — drops the per-stage layer span so worker ranks serve cross-chain pulls; §8 downstream RESTORE/ABORT ack is timeout-bounded (hang → cold).
- **dist-spec** — spec-decode KV compaction at capture (drop proposed-then-rejected positions) + draft/target depth-align.

## Review-hardening pass (22 commits on top of the rebase)

A full multi-agent review of this branch produced fixes, all applied here one-commit-per-fix:

**Correctness / silent-wrong-output class**
- Failed cold-fallback now returns `Chunk::error`, not an empty success marker.
- Every abandoned `set_state_blob` (failed or applied-then-discarded) scrubs with `recreate_request` — `reset_state` provably cannot clear post-`set_state` residue (ov-runtime + dist-spec; qwen36/gemma4 already did).
- Chain ABORT rollback scrubs effectively and error-logs failed downstream relays instead of acking clean over a swallowed failure.
- `abort_warm_resume` retracts the pulled downstream stashes, so an orphaned blob from an aborted move can no longer ride a later RESTORE via the single-slot fallback.
- Shim restore alignment hardened: Family-A alias parse searches the fused name (either alias order), positional fallback requires verbatim name equality, and the donor's element type must match the destination state's (closes the f16/bf16 reinterpretation case; `KV_CACHE_PRECISION` is per-node).

**Availability / protocol**
- Every §8 RESTORE/CAPTURE ack wait is now strictly bounded (15 s), and the connection is dropped on timeout so the late, unsequenced ack cannot pair with the next exchange (sparse-moe, dist-spec driver + relays; CaptureAck sites too).
- CAPTURE bodies are read in ≤64 KiB chunks (they outgrow the transport's single-read cap at ≈13–16k tokens), and over-ceiling carried blobs are drained before rejecting — no more frame desync on long sessions.
- **Wire change:** the single-token Forward family gained the `push_history` byte, so its frame kinds are bumped (0x04/0x06/0x07 → 0x0C/0x0D/0x0E) per the file's versioning convention — a mixed-version chain now rejects loudly instead of shifting every field by one.

**DoS**
- `ov_blob_decode` bounds its wire-declared layer count before allocating (a 12-byte blob could request ~240 GB); the shim validates declared dims/dtype against the payload length before constructing the restore tensor.
- The kv-wire alloc-DoS regression test now actually exercises the bincode byte budget (the forged varint used the wrong marker), and `LimitExceeded` maps to `FrameError::TooLarge` so a DoS attempt is distinguishable from corruption.

**Tenant isolation**
- An oversized tenant fails the capture instead of silently downgrading CaptureV2 to the world-readable untagged v1 frame.

**Capture-side hygiene**
- `get_state_blob` fails on an unmappable element type at export (was: a blob that fails on every consumer) and errors on a non-null under-sized buffer (was: silent size-query degradation).
- `cargo test -p cascadia-engine-sparse-moe` compiles on Windows again (the symlink-mirroring k26 test is now `cfg(unix)`).

## Testing

Wire contract — codec roundtrip + golden vectors:

```bash
cargo test -p cascadia-kv-wire
```

Engine KV surfaces (the feature is off by default in both engine crates):

```bash
cargo test -p cascadia-engine-openvino -p cascadia-engine-sparse-moe --features cascadia-engine-openvino/kv_coord,cascadia-engine-sparse-moe/kv_coord
```

Full workspace:

```bash
cargo test --workspace
```

- kv_coord engine suites: 478 pass (Windows) / 479 (macOS). Full workspace: 788 pass (787 on Windows — one test is deliberately `cfg(unix)`).
- `b320fac6` fixes the qwen36-moe cross-chain regression found by the #53 move cert: the dtype guard added in review hardening read the destination via `get_state()`, which throws (`Cannot get length of dynamic dimension`) on a fresh recipient's dynamic-in-T attention states — the guard now compares against a dtype cached from materialized reads and never materializes a dynamic destination; a never-materialized state applies on size+identity with a stderr notice.
- Verified on the matias fleet: kv-wire 38/0, kv_coord 478/0, workspace 787/0, plus `--features openvino` 151/0 and openvino+kv_coord 202/0 — the C++ shim changes compile and pass against real OpenVINO 2026.2.
- A cross-holder wire-contract conformance suite runs the same clauses against both MoE holders, so the two cannot drift apart.
- Rig coverage runs via enterprise #53 (cross-chain pull-HIT on the 4-node sharded rig, `warm==cold` byte-identical).

## Merge notes

- **Rebase onto `main`: done.** This branch sits on `main` (post-#138) with the review-hardening commits on top; **enterprise #53 must re-pin to the current tip `b320fac6`** (its previous pin `f678a8f3` predates the rebase + fixes).
- **Merge, do not squash.** #53 pins this branch by git rev; a squash-merge orphans that pin.
- **Mixed-build note:** the Forward frame-kind bump and the stricter shim restore guards mean old/new nodes on one chain reject loudly instead of desyncing or scrambling — upgrade chain nodes together (the rig story already assumes this).
- The 2026-08-19 rig cells (13 cells incl. the 10-cell formation soak and qwen36 `CERT PASS 7/7`) ran against enterprise builds pinning the pre-hardening `f678a8f3`; the fixes are unit/fleet-tested as above, and #53's re-pin run re-exercises the rank>0 donor path on hardware.
- Follow-ups deliberately out of this branch: an ABORT_ACK verdict byte (wire change), the dist-spec driver's inherited no-op `abort_warm_resume`, and plane-side `kv_handoff.clear(epoch)` on abort (enterprise repo).

Relates to: cascadia-enterprise#34 (parent) · cascadia-enterprise#53 (enterprise half).


7i7o (approved):
tACK 👍

PR #112, opened 2026-07-22
feat(metrics): Prometheus /metrics endpoint — request, generation, engine, and transport metrics (#16)

### LINKED ISSUE #16: Prometheus /metrics endpoint with request, generation, engine, transport, and MoE metrics

### LINKED ISSUE #124: Shutdown mid-generation: metrics book nothing, or book a cancel, depending on a race
Found during the review of #112 (Prometheus metrics). Not introduced by it — #112 is the first thing that makes the behaviour *observable*, and it needs a design call rather than a quick patch, so it is split out.

## What happens

`ChunkStream::poll_next` has a teardown arm for "the engine slot is empty" (`crates/cascadia-runner/src/lib.rs`, the `let Some(engine) = guard.as_mut() else` branch):

```rust
this.done = true;
// Server teardown (Runner::close emptied the slot), not a client cancel:
// finalize with NO terminal sample so Drop doesn't book every in-flight
// generation of a shutdown as a client cancellation.
this.metrics_finalized = true;
return Poll::Ready(None);
```

Suppressing the cancel is right — a restart is not a client cancellation. The problem is that **which** outcome a shutdown books depends on an unsynchronised race:

| ordering | booked |
|---|---|
| stream is polled after `Runner::close()` empties the slot | **nothing** |
| stream is dropped first (client hung up during the drain, or the runtime drops the connection task at exit) | **`cancelled`**, via `impl Drop for ChunkStream` |

Nothing sequences those. Per-connection tasks are `tokio::spawn`ed, so dropping the accept loop does not abort them, and both orderings are reachable on the normal CLI shutdown path.

## Why it matters

- `cascadia_tasks_cancelled_total` gains a **nondeterministic** bump on every restart, sized by whichever streams lost the race. That is worse than a consistently wrong number, because it cannot be corrected for — a rolling restart across a fleet produces cancel spikes that look like real client churn.
- In the poll-first ordering nothing is booked at all, so `completed + failed + cancelled < started` after any shutdown, and there is no metric for "aborted by shutdown". No invariant test asserts that sum anywhere.

The existing test `teardown_mid_generation_is_not_counted_cancelled` only exercises the poll-first ordering, so it green-lights a contract the code does not actually enforce.

`crates/cascadia-metrics/src/lib.rs` now documents the suppression as conditional rather than guaranteed (done in #112), but that is describing the bug, not fixing it.

## Options

1. **Make it deterministic.** Put a `closing: Arc<AtomicBool>` on `Runner`, set it at the top of `close()` *before* emptying the slot, clone it into `ChunkStream`, and check it first in `record_cancelled_metrics`. Both orderings then collapse onto one outcome, and the anomalous `metrics_finalized = true` write in the teardown arm can be deleted — that arm becomes a plain `done = true; return Ready(None)`.
2. **Give it its own terminal reason.** Observe `cascadia_generation_duration_seconds{finish_reason="teardown"}` in the teardown arm, so the sum invariant holds and "how many generations did that restart cut off?" becomes answerable. Composes with option 1.

Option 1 is the correctness fix; option 2 is the observability one. They are complementary, and 1 is net line-reducing.

Whichever is chosen, the regression test needs a drop-first case, not just the poll-first one it has today.

## Related

Closely tied to the truncated-200 issue filed alongside this one — that is the *client-visible* half of the same shutdown path.

### LINKED ISSUE #125: Shutdown mid-generation returns 200 OK with a truncated body and finish_reason "stop"
Pre-existing on `main`; found during the review of #112, which is why it is filed now rather than fixed there. This is the client-visible half of the shutdown path — see the metrics half in the companion issue.

## What happens

When `Runner::close()` empties the engine slot with a generation in flight, `ChunkStream::poll_next` takes its teardown arm and returns `Poll::Ready(None)` — a clean end-of-stream, indistinguishable from "the model finished".

The non-streaming handler in `crates/cascadia-api/src/lib.rs` initialises

```rust
let mut finish_reason: &'static str = "stop";
```

and drains with `while let Some(chunk) = chunk_stream.next().await { ... }`. On teardown that loop simply ends. The handler then builds a normal **200 OK** carrying whatever partial text had accumulated, `finish_reason: "stop"`, and a `usage` block reflecting the partial token count.

The SSE path behaves the same way: the chained tail emits the usage frame and `data: [DONE]`, so a streaming client also sees a well-formed, successful completion.

## Why it matters

The client is told the model finished normally when it did not. A caller has no way to distinguish "the model stopped here" from "the server was restarting" — the response is byte-for-byte the shape of a successful short answer. Anything that persists completions (chat history, eval harnesses, fine-tuning data collection) silently records truncated output as complete.

This is the exact failure `Chunk::error` was introduced to prevent; see the doc comment on `Chunk::error_marker` in `crates/cascadia-types/src/chunk.rs`, which makes the same argument for engine failures. Shutdown is the remaining hole.

## Suggested direction

Have the teardown arm surface a terminal error chunk rather than a bare end-of-stream, so it routes through the existing error paths (503 / SSE error frame) instead of building a 200. `fail_stream()` already does exactly this for the three other terminal exits in `poll_next` — but note it currently books `tasks_failed_total`, and a graceful restart should probably not land on the primary failure SLO, so this wants coordinating with the companion metrics issue rather than reusing `fail_stream` as-is.

A `finish_reason` distinct from `stop` (OpenAI has no standard value for this; `cancelled` is the closest existing one in our enum) would at minimum stop the response *claiming* a normal stop.

## Notes

- Reachable on the normal CLI shutdown path: per-connection tasks are `tokio::spawn`ed, so they keep running after the accept loop is dropped and are still being polled while `close()` runs.
- No test covers the client-visible side of teardown today; the one teardown test asserts only that no cancel is booked.

### LINKED ISSUE #126: ov-genai (non-CB) turns a generate() failure into an empty 200, so tasks_failed_total stays 0 on the default engine
Pre-existing on `main`; found during the review of #112. Out of scope there (that PR makes zero engine changes), but it directly undermines a metric #112 ships, so it should be fixed before anyone trusts that metric.

## What happens

`crates/cascadia-engine-openvino/src/genai.rs`, in the non-CB `GenaiEngine::step()`:

```rust
let result = match self.pipe.generate(&task.prompt, &cfg) {
    Ok(r) => r,
    Err(err) => {
        warn!(task = %task.task_id, error = %err, "generate failed");
        let final_chunk = Chunk::final_marker(task.task_id.clone(), "");
        return Ok(vec![(task.task_id, final_chunk)]);
    }
};
```

A failed generation returns `Ok(...)` carrying an **empty successful final marker** instead of `Chunk::error(...)`. The failure exists only as a `warn!` line.

## Two consequences

**1. The client gets a 200 with empty content.** `Chunk::final_marker` sets `error: None` and `finish_reason: None`, so the API builds a normal success response with an empty message. This is precisely the case `Chunk::error` was added to prevent — see the doc comment on `Chunk::error_marker` in `crates/cascadia-types/src/chunk.rs`: a bare end-of-stream "builds a 200 with truncated text and finish_reason stop at the API layer", which is indistinguishable from "the model said nothing".

**2. `cascadia_tasks_failed_total` cannot fire here.** The runner's `record_chunk_metrics` books `tasks_failed_total` only when `chunk.error.is_some()`. An empty final marker is instead booked as a **successful** generation with `finish_reason="stop"` and zero tokens.

That matters more than it looks, because of what the default is: `--engine` defaults to `ov-genai` and `--cb` defaults to false, so this is the default worker configuration. On a default deployment, the most common engine-side failure is invisible to the failure metric and shows up as a *successful* zero-token generation.

The observable signature today is a rise in `cascadia_generation_duration_seconds{finish_reason="stop"}` with `cascadia_tokens_generated_total` flat — which nobody will alert on.

## Fix

Return `Chunk::error(task.task_id.clone(), err.to_string())` instead of `final_marker`. `Chunk::error` already sets `is_final`, so the shape of the return is unchanged; it routes through the existing loud paths (503 / SSE error frame) and books `tasks_failed_total`.

Worth auditing the sibling engines for the same pattern while in there.

## Verification

Needs Intel hardware with `--features openvino` to exercise the real failure, but the accounting half is unit-testable through the existing `CbPipe`-style seam: a stub whose `generate` fails should produce an error chunk, and the runner should book `tasks_failed_total` rather than `duration{finish_reason="stop"}`.

t8:
Closes #16. Closes #124. Closes #125. Closes #126.

## What

Adds production observability: every stage started with `--api` now serves Prometheus text exposition at `GET /metrics`. Before this, the codebase had zero metrics (only `tracing` logs).

A new leaf crate **`cascadia-metrics`** holds the process-global registry (`prometheus = 0.13`, `default-features = false` so the protobuf exposition dep stays out of the tree; passes `cargo deny check all`). The three emitting layers depend on it — no dependency cycles, and **zero per-engine changes**:

| Layer | Metrics |
|---|---|
| `cascadia-api` | `http_requests_total{endpoint,status}`, `http_request_duration_seconds{endpoint}`, `inflight_tasks`, `api_rejected_total{reason}` |
| `cascadia-runner` | `generation_ttft_seconds{model}`, `generation_inter_token_seconds{model}`, `generation_duration_seconds{model,finish_reason}`, `tokens_generated_total`, `tokens_prompt_total`, `tasks_cancelled_total`, `tasks_failed_total`, `engine_model_load_duration_seconds{model,device}`, `engine_warmup_duration_seconds{model,device}` |
| `cascadia-transport` | `transport_sent_bytes_total{kind}`, `transport_recv_bytes_total{kind}`, `transport_send_seconds`, `transport_recv_payload_seconds` |

All names use the `cascadia_` prefix (the issue predates the tahoma→cascadia rename). Buckets follow the layouts proposed in #16. Full inventory + example PromQL in **docs/METRICS.md**.

## Design notes

- **Bounded label cardinality by construction**: `endpoint` is the *matched route template* (`/v1/cancel/:task_id`, never the raw URI, verified live), `model`/`device` are fixed per process, `reason`/`finish_reason` are closed enums.
- **Generation metrics live at the runner's `ChunkStream`** — the single choke point every chunk flows through — so streaming + non-streaming on both `/v1/chat/completions` and `/v1/completions` and every engine are covered by one instrumentation site. Terminal accounting is exactly-once (`metrics_finalized` guard): completed → `finish_reason` label, engine error → `tasks_failed_total`, explicit cancel or client disconnect mid-stream → `tasks_cancelled_total` (via `Drop`).
- **Transport recv histogram excludes the frame-start wait** (idle stages legitimately sit in it between requests); it measures header-complete → payload-complete. Byte counters include headers and cover both tensor and raw (dist-spec control) frames.
- No queued-tasks gauge: admission is `try_acquire` → immediate 503, there is no queue. `api_rejected_total{reason="capacity"}` is the pressure signal.

## Deviations from the issue

- `tahoma_*` → `cascadia_*` naming (repo rename).
- **Sparse-MoE per-layer metrics deferred** (the issue flags them as extra scope): the hook point is identified (`dispatch_expert` in `sparse-moe/src/runner.rs`) but wiring the hot kernel path deserves its own change validated on real MoE hardware. Tracked in docs/METRICS.md follow-ups.
- Relay-only stages (no `--api`) expose no scrape endpoint today — documented; a metrics-only listener is a follow-up.

## Self-review round (second commit)

A multi-angle self-review (3 correctness finders + cleanup/altitude finders, adversarial verify) caught four issues, fixed in the second commit: engine-acknowledged cancels (`FinishReason::Cancelled` final marker) now increment `tasks_cancelled_total` so the two cancelled series can't diverge; server teardown with in-flight generations is no longer booked as client cancellations; engine-queue-full 503s (`EngineError::QueueFull`, reachable at default config since engines cap pending below the permit count) now count as `capacity` rejections; and empty final markers no longer pollute TTFT/inter-token histograms. Same round moved the #55 token-count convention to `Chunk::token_count()` in `cascadia-types` (single source for API usage, dashboard, and metrics), collapsed the three error-return sites in `ChunkStream::poll_next` into a `fail_stream()` helper so terminal accounting is structural, and tightened the measurement-semantics documentation (delivery-time timing, SSE 200-head status caveat, layer-level rejections, dashboard-route scope, transport full-frames-only accounting). Each fix has a dedicated test.

## Testing

- `cascadia-metrics` unit tests (registration, exposition format, bucket layouts).
- `cascadia-api` integration tests through the mock engine: exposition content-type, request/TTFT/duration/token/engine-gauge families populated by a real request (isolated by dedicated model label), rejection counter.
- `cascadia-runner` tests: token/TTFT/duration accounting on completion; cancelled counted exactly once on drop; completed streams NOT counted cancelled; failed generations counted failed and not double-counted as cancelled.
- `cascadia-transport` test: roundtrip advances byte counters + duration samples.
- e2e (`tests-e2e`): the spawned `cascadia` binary serves `/metrics`; chat request populates counters; `cascadia_inflight_tasks` returns to 0.
- Manual live run: `cascadia worker --engine mock --api`, non-streaming + streaming + cancel + rejection traffic, scrape verified (matched-route labels, finish_reason length/stop split, token counts exact).
- `cargo fmt --check`, `clippy` (no new warnings in touched crates), `cargo test --workspace --all-targets` (49/49 suites green), `cargo deny check all` — all pass locally.

All testing is local-only (loopback + mock engine); no fleet nodes were touched.


t8:
Note on the pending `build + test (openvino feature, ai-pc)` check: the self-hosted Windows runner is not picking up jobs — the run for `pawan/deepseek-v4-sparse-moe` queued ~12h before this PR is also still queued (run 29849907456), and #105 merged with its run in the same state. Both hosted checks (cargo test, cargo-deny) are green on the final commit. The changes here are feature-independent (no openvino-gated code touched), so the ai-pc job should pass whenever the runner comes back.

7i7o:
Follow-ups split out of this PR's review, all pre-existing on `main` rather than introduced here:

- #124 — shutdown mid-generation books nothing, or books a cancel, depending on a race (needs a design call)
- #125 — shutdown mid-generation returns 200 OK with a truncated body and `finish_reason: "stop"`
- #126 — non-CB `ov-genai` turns a `generate()` failure into an empty 200, so `tasks_failed_total` stays 0 on the default engine

#126 in particular is worth doing before anyone relies on `cascadia_tasks_failed_total`.

7i7o (approved):
tACK 👍

PR #114, opened 2026-07-23
feat: glm-5.2 support in the sparse-moe engine

pawanpaudel93:
## Summary

Adds end-to-end support for **GLM-5.2** (`glm5`, ~744B sparse-MoE, int4 experts) to the sparse-moe engine: an exporter, a CPU reference model, the Rust shell, and a distributed pipeline that shards the model one stage per node across 4× 32 GB Intel AI-PCs.

GLM-5.2 uses DeepSeek-V3-style **MLA latent attention** plus a **DSA sparse indexer**, so the shell reuses the arch-agnostic `PipelineEngine<R>` introduced for dsv4 — most of the runtime work (batched prefill, KV-prefix cache, per-token streaming) lands for **both** engines.

## What's included

- **Exporter** (`tools/export_glm5.py`) — FP8 → int4 per-expert repack + shell safetensors, a `--tiny` synthetic mode for the correctness harness, and the serving sidecars (tokenizer + `chat_template.jinja`) so chat completions render correctly.
- **CPU reference** (`tools/glm5_ref/`) — a deterministic pure-torch reference (MLA, DSA indexer, IndexShare, sparse-MoE) that generates the golden fixtures the Rust shell is tested against.
- **Rust shell** (`crates/cascadia-engine-sparse-moe/src/glm/`) — MLA latent attention, the DSA indexer with **IndexShare** (shared layers reuse the carried top-k), sparse-MoE experts as mmap int4 (default) or optional per-expert OpenVINO IR, and the MTP draft head. Mirrors the CPU reference 1:1.
- **Distributed pipeline** (`src/engine.rs`, `src/dist.rs`) — one stage per node; rank 0 tokenizes + embeds + drives, mid ranks relay the hidden state, the last rank runs the head + sampler. Token ids never leave rank 0.
- **Batched prefill** (`ForwardBatchPrefill`) — the prompt runs as batch-union-MoE frame(s) per rank (≤256 rows each; longer prompts chunk into windows — see *Decode performance & long context*), validated **bit-exact** vs the single-process prefill across 1/2/4 ranks.
- **Distributed KV-prefix cache** (opt-in, `CASCADIA_GLM5_PREFIX_CACHE`) — snapshot/restore the MLA + DSA KV keyed by prompt-token prefix; on a hit, restore the longest cached prefix on every rank and prefill only the suffix. Caches the full `[prompt + response]` KV post-decode, so the next turn reuses the whole conversation (prompt **and** response). Bit-exact parity with a full prefill across 1/2/4 shards.
- **Per-token streaming** — `PipelineEngine::step` emits one SSE `Chunk::token` (UTF-8-safe delta) per step instead of returning the whole completion on one final marker, so the client sees the first token at TTFT rather than after the full decode. Covers **glm5 + dsv4**.
- **OpenVINO expert backend** (opt-in, `CASCADIA_GLM5_OV_EXPERTS`) — runs each routed / shared expert as a compiled OV int4 IR on **iGPU / NPU / CPU** instead of the Rust int4 kernel (`glm/ov_expert.rs` + exporter `tools/glm5_expert_ov.py`). A port of the dsv4 backend — **experimental**; see [iGPU / NPU expert offload](#igpu--npu-expert-offload-scale-out) for the measured caveats (dsv4 fleet: slower than the Rust path).
- **Async lookahead expert prefetch** (opt-in, `CASCADIA_GLM5_LOOKAHEAD`, **off by default**) — a background thread warms the next layer's predicted, non-resident experts into the OS page cache while the main thread computes. Prediction is the next layer's own router on an attention-free proxy of the previous layer's output; **miss-only and output-identical** (the warm never feeds compute). Measured **~9% lower decode compute/token** on the 4-node fleet (**not independently re-verified**; the figure comes from the same profiler whose hit% is understated — see above); stays opt-in because the end-to-end gain is small and dominated by inter-node network variance (see *Feature flags* + *Validation*).
- **Windows autopin fix** (always on) — the hot-expert `mlock` now actually sticks on Windows: available RAM is read via `GlobalMemoryStatusEx` (was a hardcoded 8 GB fallback → pin budget computed to ~0), and the process working set is raised via `SetProcessWorkingSetSizeEx` before `VirtualLock` (which otherwise fails the working-set quota and pinned **0/N** experts). Pure correctness — no throughput change on the current 32 GB fleet (the budget covers only a few % of a rank's experts under near-uniform routing), but it matters once a node holds a larger expert fraction.
- **Also**: seeded temperature parity (RNG draw discipline matches single-stage), grammar-constrained decoding, and a per-section decode profiler (`CASCADIA_GLM5_PROFILE`).

### Decode performance & long context

- **bf16 embed / lm_head** (**engine default**; opt out with `CASCADIA_GLM5_F32_HEAD`) — stores the edge-rank embed + lm_head as bf16, freeing ~3 GB/edge-rank back into the expert pin budget. Measured **~1.2× end-to-end decode** (fixes the RAM-starved `lm_head` rank, which was ~2× slower than the others); logits stay within bf16 tolerance of f32 (the attention shells are already bf16, so f32 embed/lm_head was the anomaly). Precision is an explicit `StageOpts { bf16_head, bf16_kv }` threaded from the serving entry into the loader — the env is interpreted in **one place** (`StageOpts::from_env`), so the pin-budget reservation and the KV/head allocations derive from the same value.
- **Chunked prefill** — a prompt longer than `MAX_BATCH_COUNT` (256) is now prefilled as a sequence of ≤256-row windows (new `ForwardBatchPrefillNoSample` frame). KV accumulates across windows exactly as token-by-token and only the final window samples, so no mid-prompt row is recorded into rep-penalty. Prompts >256 tokens previously errored (`batch_count out of range`); long-context input now works over the pipeline.
- **bf16 MLA KV cache** (opt-in, `CASCADIA_GLM5_BF16_KV`) — stores the MLA latent / k_pe cache as bf16, halving KV RAM (∝ `max_seq`; ~0.7–3 GB/rank at 32k–128k). Long-context recall validated identical to f32 (2-needle eval); off by default (no normal-context benefit, and it changes free-gen output).
- **Residency telemetry** — `CASCADIA_GLM5_PROFILE` now also reports, per token per rank: bytes streamed at 0%-hit, the **true expert-cache hit %** (probed from the working set via `QueryWorkingSetEx`/`mincore` — the process read counter misses mmap paging on Windows), achieved expert MB/s, and cross-token expert reuse %.

## Performance

GLM-5.2 decode is **I/O-bound**: at 744B the routed experts do not fit in RAM, so each token streams its experts from NVMe. The profiler (`CASCADIA_GLM5_PROFILE`) attributes steady-state decode to **~78% expert streaming, ~20% attention, ~0.5% router** — the model is waiting on disk, not compute. The runtime levers target that wall:

- **bf16 embed / lm_head** (default, **~1.2×**) frees edge-rank RAM into the pin budget → more experts resident. The working-set probe shows steady-state expert-cache **hit % ≈ 64–73%** across ranks (**understated by an unknown amount** — the probe tests `PSAPI_WORKING_SET_EX_BLOCK.Valid`, i.e. "in this process's working set", while `trim_working_set()` deliberately moves pages to the standby list where they remain in RAM but read `Valid=0`; treat as a lower bound until the probe decodes the `Invalid` block's `Location` field), so the engine is **residency-bound**: freeing RAM — or sharding wider — is the throughput lever, not faster kernels.
- **KV-prefix cache** — reusing a shared conversation prefix skips re-streaming that prefix's experts, the dominant cost. Measured **~2.15× faster TTFT** on the agentic multi-turn pattern (warm prefix hit) on the 4-node fleet.
- **Per-token streaming** — orthogonal to raw throughput but the decisive UX lever on a slow-decode path: at seconds-per-token, a client that waits for the *whole* completion is unusable for agentic/interactive work. Streaming surfaces the first token immediately and every token as it lands, and enables mid-stream cancellation.

### Feature flags & defaults

**Default path (no env set):** two things are **on** — **bf16 embed/lm_head** (~1.2× decode, logits within bf16 tolerance; opt out with `CASCADIA_GLM5_F32_HEAD=1` to force exact f32) and the **Windows hot-expert autopin fix** (correctness; no flag). **Every flag in the table below is OFF unless you set it** — all opt-in, all read from the environment at the serving entry.

| env (opt-in) | what it does | why it's off by default | to enable |
|---|---|---|---|
| `CASCADIA_GLM5_PREFIX_CACHE` | reuse cached conversation-prefix KV | costs RAM; helps only multi-turn TTFT (measured ~2.15×) | `=<depth>` (e.g. `8`) |
| `CASCADIA_GLM5_BF16_KV` | bf16 MLA latent / k_pe KV cache | changes free-gen output; frees RAM only at long context | `=1` (long-context) |
| `CASCADIA_GLM5_LOOKAHEAD` | async lookahead expert prefetch (one background reader thread) | lossless ~9% compute win, but small + network-dominated end-to-end (n-small) | `=1` |
| `CASCADIA_GLM5_OV_EXPERTS` | run experts on iGPU / NPU / CPU via OpenVINO IR | experimental; measured *slower* than the Rust kernel until experts are RAM-resident | `=1` **and** a `<model>/experts_ov/` dir |
| `CASCADIA_GLM5_PROFILE` | per-token decode profiler → stderr | measurement only (stderr noise) | `=1` when measuring |
| `CASCADIA_GLM5_PREFETCH` | histogram `WILLNEED` next-layer prefetch | measured *slower* under memory pressure (readahead re-evicted) | `=1` (not recommended) |
| `CASCADIA_GLM5_R1READ` | concurrent whole-bin expert reads | evicts the page cache on the fleet (hit % → ~50%) | `=1` (not recommended) |

> **Flag values are now parsed, not just detected.** These switches previously
> tested presence only, so `FLAG=0` — the obvious way to turn one off — switched it
> ON. `glm::env_flag` now treats unset / empty / `0` / `false` / `no` / `off` as off
> and anything else as on, across all 12 `CASCADIA_*` gates in the engine. This
> mattered most for `CASCADIA_GLM5_OV_EXPERTS=0` (would have opted into the
> experimental OpenVINO backend) and `CASCADIA_GLM5_F32_HEAD=0` (would have forced
> the f32 head, changing the pin budget).

**Recommended to enable** (the default path — bf16 head + autopin — is already the safe baseline; add these per workload):

- **Agentic / multi-turn** → `CASCADIA_GLM5_PREFIX_CACHE=8` — ~2.15× TTFT, bit-exact. The single highest-value flag for conversational use.
- **Long context (≥ 32k)** → also `CASCADIA_GLM5_BF16_KV=1` — frees KV RAM (∝ `max_seq`), recall-validated. Skip at normal context.
- **Optional** → `CASCADIA_GLM5_LOOKAHEAD=1` — lossless ~9% decode compute; A/B it on your traffic (small, network-swamped end-to-end).

Everything else stays off. Example agentic launch:

```bash
CASCADIA_GLM5_PREFIX_CACHE=8 \
CASCADIA_GLM5_BF16_KV=1 \        # long-context deploys only
  cascadia worker …             # bf16 head + autopin already on by default
```

Non-feature config envs, not shown above: `CASCADIA_GLM5_MAX_SEQ` (context budget, default 4096), `CASCADIA_GLM5_EXPERTS` (expert storage mode, default mmap int4), `CASCADIA_GLM5_NOPIN_ACTIVE` (disables the always-active shared/dense pin — debug / A-B), and the OpenVINO sub-options `CASCADIA_GLM5_OV_DEVICE` (default `GPU`) / `OV_CACHE` / `OV_CACHE_DIR` / `OV_STATS`, which apply only when `OV_EXPERTS` is set.

### iGPU / NPU expert offload (scale-out)

**Experimental, and measured to *lose* on the sibling engine — off by default.** Because decode is I/O-bound, moving the expert GEMV to the iGPU/NPU only helps once a rank's experts are **RAM-resident** — on Lunar Lake the CPU/iGPU/NPU share one unified system RAM, so a shard that streams from NVMe just moves the same wait onto the accelerator. Even RAM-resident, the dsv4 fleet measured the end-to-end distributed OV path at **~0.6 tok/s vs ~1.05 tok/s for the optimized Rust kernel** — *slower* and fragile (the per-expert compile cache thrashes once the working set exceeds the LRU cap / RAM). A single-expert microbenchmark favors OV (~4.6× CPU / ~6× iGPU, batch=1), but that did not survive the real workload. Kept as an opt-in option; **the Rust int4 path is the default and the one to prefer.**

The backend is built and **dormant by default**:

- **Runtime** (`glm/ov_expert.rs`) — runs each routed/shared expert as a compiled OV IR; `None` (Rust int4 path) unless `CASCADIA_GLM5_OV_EXPERTS=1` **and** `<model>/experts_ov/` exists. A missing/uncompilable IR falls back to the Rust kernel (warn once) rather than taking down a rank.
- **Exporter** (`tools/glm5_expert_ov.py`, requires `nncf`) — converts the int4 expert bins into per-expert OV IRs via **NNCF `INT4_SYM`** weight-compression (the same proven path as the dsv4 fleet tool), so the resident footprint matches the int4 bins (~0.56 B/weight). `--validate` compiles fp16 & int4 and compares to a numpy reference; `--dtype fp16` is a spare-RAM/debug escape hatch (~3.6× larger). The dsv4 counterpart (`tools/dsv4_expert_ov.py`) is vendored in-repo too, resolving the previously-dangling `dsv4/ov_expert.rs` reference.
- **Enable**: `python tools/glm5_expert_ov.py --src glm5-export` → then run with `CASCADIA_GLM5_OV_EXPERTS=1 CASCADIA_GLM5_OV_DEVICE=GPU` (or `NPU` / `CPU`).

> Status: the runtime is a no-op-tested port of the dsv4 backend (multi-agent-reviewed; the confirmed defects are fixed); the glm IR export + on-device compile/infer are **unvalidated in CI** (no OpenVINO/checkpoint). Given dsv4's measured result, treat this as experimental, not a throughput win.

## Validation

- FP8 → int4 dequant/repack golden-tested against the reference on real shard math.
- Rust shell matches the CPU reference greedy on the `--tiny`/8-layer harnesses (real structural dims): MLA, DSA indexer, IndexShare, MoE gate + experts, RoPE, layer, model.
- 1-, 2-, and 4-rank loopback chains over real `cascadia-transport` match the single-process reference (batched prefill **bit-exact**); mmap and eager expert dispatch agree.
- **Distributed KV-prefix cache**: caching a base prefix, restoring it, and prefilling only the suffix is **bit-identical** to a full prefill across 1/2/4 shards (`glm5_prefix_cache_pipeline_matches_full`).
- **Streaming (live, 4-node fleet):** a streaming `/v1/chat/completions` returns `content-type: text/event-stream` with **time-separated per-token frames** (each token delivered as generated, not batched at the end), coherent output, and correct finalization (`finish_reason` → `[DONE]`). UTF-8-safe delta slicing is unit-tested for multi-byte splits.
- **bf16-head path covered in CI**: `glm5_bf16_head_pipeline_bounded_vs_f32` runs the default bf16 embed/lm_head through the staged pipeline and bounds its logits against the exact-f32 reference (bounded error within the model's bf16 contract — not brittle byte-identity). `WideTable` / `KvStore` unit tests bound the bf16 head/KV math directly.
- The full workspace test suite passes; `cascadia-api` + `cascadia-cli` build clean.

### On-fleet validation (4× Lunar Lake, CPU)

Ran the test plan below; all green. Tier A (local gate) re-verified at the current head **`bd4df50`**; the Tier B chat / streaming / long-context captures are from `6f64995` and are unaffected by the `bd4df50` additions (the lookahead prefetch is opt-in / default-off and the autopin fix is Windows correctness — neither changes chat output); the async-prefetch A/B/C + autopin check were run at `bd4df50`.

**Tier A (local, no model/hardware) — HEAD `bd4df50`:** `cargo fmt --all --check` ✓ · clippy (sparse-moe) ✓ · **180 tests, 0 skip / 0 fail** — every gated `glm5_*` test ran (1/2/4-rank wire protocol, prefix-cache parity, bf16-head bounded-vs-f32, the lookahead router-prediction unit test, IndexShare, MLA/DSA, MoE, UTF-8 streaming) ✓ · exporter `--tiny` smoke ✓.

**Tier B (4-node fleet):**

- **B2 build** — all 4 nodes built clean at the branch head.
- **B3 chat** — `List three primary colors` → coherent `Red / Blue / Yellow`; chain torn down cleanly.
- **B4 streaming** — `content-type: text/event-stream`, **time-separated** per-token frames (Red @124 s → Blue @140 s → Yellow @145 s, seconds apart = true per-token streaming), a `finish_reason: stop` marker, then `data: [DONE]`.
- **Long context (chunked prefill)** — a ~370-token, 2-window prompt prefilled with **no `batch_count` cap error** and correctly recalled a needle planted in window 1 (`ORANGE-7392-DEL…` — the exact prefix, truncated by the eval's `max_tokens`); the working-set profiler reported expert-cache **hit ≈ 50–55%** on the decode tokens (lower than the ~64–73% at short context — a 2-window prompt spreads routing across more experts).
- **iGPU/NPU offload (B5)** — not run; experimental and unvalidated in CI (see the offload section), and measured slower than the Rust kernel on the sibling engine.

**Async lookahead prefetch — A/B/C (HEAD `bd4df50`).** Baseline vs `CASCADIA_GLM5_R1READ` vs `CASCADIA_GLM5_LOOKAHEAD`, per-token-normalized (via the API's `completion_tokens`) with a **compute-vs-network-gap split**, warmth-controlled (run in both pass-orders). `LOOKAHEAD` cut per-rank decode **compute ~9%** — **warmth-independent, output-identical, recall 80–91%** — at *flat* expert-cache hit%, i.e. latency-hiding (soft-faults), not higher residency. End-to-end moved **~+8% tok/s when the inter-node network was stable**, but the network-gap term varies run-to-run and often dominates, so the feature ships **opt-in**. `R1READ` **lost** (whole-bin reads evict the page cache → hit% ~50%). The **Windows autopin fix** was verified live — hot-expert pins reported **`205/205`** where the unfixed build pinned **`0/N`**.

The `cascadia-fleet-tests` harness is idempotent (per-node kill-first + port-matched teardown), so every step is safely re-runnable.

## Post-review fixes

17 commits on top of the original branch, from a multi-agent review plus a full
Tier A + Tier B run on a 4-node Lunar Lake fleet. Grouped by what they change.

**Behaviour changes worth knowing when rebasing onto this**

- `prefix_remember` now returns the **existing** key for an already-indexed token
  sequence instead of minting a fresh one. The index dedups by tokens while every
  rank's `SliceKvCache` dedups by key, so a fresh key on a repeat meant the two
  sides evicted different entries and `prefix_reuse` could hand out a key no rank
  still held — which reached a release-mode `assert_eq!(base, self.pos)`.
- `finalize_pipeline` no longer caches the KV prefix when finalizing a **failed**
  decode. The runner's KV advances before the wire op while `a.pos` advances only
  on success, so the stored snapshot was one position longer than its own key.
- `restore_prefix`'s return value is now honoured on both call sites; a miss or a
  length disagreement falls back to a full prefill (and resets, since a hit has
  already moved `pos`). Workers latch and error rather than continuing at a
  position rank 0 is not at.
- `recv_token_reply` now **drops the connection** on timeout. Timing out does not
  cancel downstream work: the peer is usually alive and still computing, and its
  `Token` — 8 raw bytes, no sequence number — would otherwise be read as the
  *next* request's reply. Reproduced on the fleet: a 386-token prefill timed out
  while all four ranks were demonstrably alive.
- `prompt_tokens` is now reported instead of always `0` (the API reads
  `chunk.prompt_tokens`; this engine never set it, unlike the OV engines).

**Correctness guards**

- OV expert output whose length disagrees with `dim` is rejected rather than
  `zip`-truncated into the residual stream and counted a cache hit.
- The exporter refuses fp8 weights with no recognized block scale instead of
  writing raw e4m3 codes as weights (dequant was gated on the scale *key name*).
- `examples/nvme_readbench.rs` is gated to unix — it imported
  `std::os::unix::fs::FileExt` unconditionally, so `--all-targets` did not compile
  on Windows, which broke `cargo test` and `cargo clippy` (two of `just check`'s
  three steps) on the deployment OS.

**Tests**

- `assert_close` now fails on non-finite values. It compared with
  `d > atol + rtol*|w|`, which is false for NaN, so the element was skipped and the
  assert passed — an all-NaN kernel output passed every element-wise golden.
- New multi-window prefill test over the real frames (300-token prompt, M ∈ {1,2,4},
  `ForwardBatchPrefillNoSample` for non-final windows), verified to fail when `base`
  stops accumulating.
- New `SliceKvCache` test pinning the key-reuse/MRU eviction contract.

## On-fleet validation (4× Lunar Lake, CPU) — independent re-run

Re-run end to end on a second fleet after staging the shards there. Activations
relayed rather than on a direct link, so **absolute latencies are worse than the
figures above and are not comparable**; pass/fail and ratios are.

| check | result |
|---|---|
| Tier A `just check` **on Windows** | fmt + clippy + **712 tests** pass |
| Tier A goldens vs CPU reference | regenerated **bit-identical** to the committed fixtures |
| Tier A exporter `--tiny` | pass |
| ai-pc OV-gated suite (real OpenVINO) | **150 passed / 0 failed** |
| B3 chat | correct answer, natural stop (`Paris`) |
| B4 streaming | `text/event-stream`, frames 8–14 s apart, `finish_reason` + `[DONE]` |
| B5 prefix cache | **5.5× TTFT** multi-turn (1019 s → 186 s), `HIT reused=174 prompt=184 suffix=10` |
| B6 long context | 386-token prompt = 2 windows, needle from window 1 recalled |
| B7 / B8 | not run |

Correctness spot-checks, all clean: `Paris` (factual, natural stop),
`ORANGE-7392-DELTA` and `VIOLET-551…` (needles recalled across a 2-window prefill),
and a needle recalled from **cache-restored** KV. No garbage, repetition loops or
UTF-8 damage in any run.

Measured there: TTFT ~300 s short / ~1960 s long (386 tok), decode ~11 s/token
(~0.09 tok/s) independent of prompt length. Prefill is ~98% of a long-prompt
request, which is why prefix reuse and residency dominate over kernel speed.

Two operational notes: the default prefill reply budget
(`min(10 × recv_timeout, frame_idle_ceiling − margin)` = 600 s) is **too small for a
256-row window on a slower interconnect** — raise
`CASCADIA_ACTIVATION_TIMEOUT_SECS` / `CASCADIA_FRAME_IDLE_CEILING_SECS`; and
`cargo build` never compiles examples, so a fleet build cannot catch a
Windows-only example break the way `just check` does.

## Known limitations / future work

- **Decode is NVMe-bound (~0.1–0.2 tok/s)** — the model doesn't fit in RAM, so experts stream from disk each token. The KV-prefix cache attacks TTFT (shared prefixes); raw decode throughput needs **RAM residency**. bf16 embed/lm_head (default) is one applied step (~1.2×), but the only order-of-magnitude move is to **shard wider** (**~21–23 nodes** fully fits 744B int4: 409 GB routed experts / 439 GB full export, ~19.75 GB pinnable per 32 GB node — the earlier 377 GB figure omitted the group-32 bf16 scales, see the size table in `docs/architectures/glm5.md`), which is lossless. Calibrated low-bit experts are a further (~1.5×), quality-gated software lever; the [iGPU/NPU expert offload](#igpu--npu-expert-offload-scale-out) is built + opt-in but **experimental** — dsv4 measured it *slower* than the optimized Rust kernel at 4 nodes (compile-cache thrash while not RAM-resident), so it only becomes a candidate once sharded wide enough to be resident, and is not a throughput win today.
- **Inter-node network is the dominant *variable* decode term** — per-token latency splits into ~compute plus a ~30% pipeline gap (activation hops between physical nodes); that gap varies run-to-run and occasionally spikes, and rare transient connection resets during batched prefill force a supervisor chain-rebuild. Transport stability + shrinking that gap is the next lever — larger than the compute-side prefetch above.
- **Single in-flight request** — the pipeline serves one request, one token at a time; concurrent requests serialize (no cross-request batching).
- **`step_single_stage` (1-box) stays batch** — these 744B models run multi-node, so the single-box path is not on the streaming state machine.
- **Recovery is supervisor-driven, and incomplete in practice** — a dead peer tears the stage down and an external watchdog restarts the chain; no in-process rank-0 reconnect/replay yet. Observed on a 4-node run: after the inter-rank links dropped (idle relayed TCP), all four processes stayed **alive** with every worker hot-looping `worker recv_kind failed: not connected` every 200 ms indefinitely. Workers never exit, so an exit-code watchdog never fires; only rank 0 signals anything (`/health` → `degraded`). A chain in this state needs a manual restart.
- **Over-budget prompts truncate the tail** (the newest turn) — logged loudly, but a budget-aware trim or an explicit 4xx would be better.

## Test plan

**Which tier do you need?**
- **Tier A** — verifies *all* the correctness (shell math, quantization, int4 experts, the multi-rank wire protocol, the distributed KV-prefix cache) with **no model and no hardware**. Enough for most reviewers.
- **Tier B** — the full 4-node distributed run, for anyone with the hardware.

---

### Tier A · correctness — no model, no hardware

Run from the repo root.

**A1 — full gate** (fmt + clippy + every workspace test). Includes the 1/2/4-rank wire protocol over real `cascadia-transport` — the *same* distributed code path, single-process on a tiny model — and the distributed KV-prefix-cache parity test:

```bash
just check
```

**A2 — goldens vs the CPU reference** *(optional; Python + torch, CPU-only)*. Regenerate the gitignored fixtures, then the gated `glm5_*` tests run (prefill parity, prefix-cache parity, IndexShare, MLA/DSA, MoE):

```bash
python tools/glm5_ref/gen_fixtures.py \
  --out crates/cascadia-engine-sparse-moe/tests/fixtures/glm5
```

**A3 — exporter smoke** *(optional; torch, no download)*. FP8 → int4 repack → manifest on a synthetic model:

```bash
python tools/export_glm5.py --tiny /tmp/glm5_tiny
```

---

### Tier B · full distributed run — 4 nodes

Uses the `cascadia-fleet-tests` harness (a **separate private repo** — the `run-chain.sh` / `sync-build.sh` / `glm5-stream-check.sh` / `glm5-prefix-cache-check.sh` / `glm5-longctx-eval.sh` / `glm5-kv-quality-eval.sh` / `glm5-abc-check.sh` / `nodes.conf` drivers below live there, not in this checkout). **You need:** 4 worker nodes, 1 big-RAM export host, and the GLM-5.2 checkpoint.

**B0 · configure your nodes** — edit `nodes.conf` (SSH creds + one line per node, line order = rank 0 → 3). Replace the names/IPs/paths with your own:

```bash
SSH_USER=<your-ssh-user>
SSH_KEY="$HOME/.ssh/<your-key>"
NODES=(
  # name | ip | repo_path_on_node | openvino_dir_on_node
  "node1|100.x.x.x|C:\\path\\to\\cascadia|C:\\opt\\openvino"
  "node2|100.x.x.x|C:\\path\\to\\cascadia|C:\\opt\\openvino"
  "node3|100.x.x.x|C:\\path\\to\\cascadia|C:\\opt\\openvino"
  "node4|100.x.x.x|C:\\path\\to\\cascadia|C:\\opt\\openvino"
)
```

**B1 · stage the model** `[orchestrator]` — the int4 export already lives on the Models Shared Drive, so no local export or hand-copy is needed: `model-distribute.sh` fetches each rank's slice straight to its node. `--ranges` pins the 18/20/20/20 split the 4-node chain expects, and rclone plus the scoped Drive credential are already provisioned on the fleet:

```bash
./model-distribute.sh glm5 node1,node2,node3,node4 --ranges 0-18,18-38,38-58,58-78 --survey
./model-distribute.sh glm5 node1,node2,node3,node4 --ranges 0-18,18-38,38-58,58-78
./model-distribute.sh glm5 node1,node2,node3,node4 --ranges 0-18,18-38,38-58,58-78 --verify
```

`--survey` reports the gap set without transferring, `--verify` re-checks by MD5 against Drive rather than trusting a size match. Idempotent — a re-run fetches only the gaps, so an interrupted transfer resumes. Lands in `C:\cascadia\models\glm5`, one rank slice per node in rank order.

<details>
<summary>Regenerating the export from the original checkpoint (rarely needed)</summary>

Only when the int4 export itself must be rebuilt — FP8 → int4 per-expert repack + shell/sidecars:

```bash
python tools/export_glm5.py --model ./GLM-5.2 --out glm5-export
```

Then either copy each rank's stage to `C:\cascadia\models\glm5` by hand, or upload the export to the Shared Drive so `model-distribute.sh` can serve it to the fleet.
</details>

**B2 · build on every node** `[orchestrator]`:

```bash
./sync-build.sh pawan/glm-5-2
```

**B3 · run + query (chat)** `[orchestrator]` — `node1,…,node4` are your `nodes.conf` names in rank order. The KV-prefix cache is forwarded here via `CASCADIA_GLM5_PREFIX_CACHE=8` — **the recommended setting for agentic / multi-turn** (see *Feature flags*):

```bash
CASCADIA_GLM5_PREFIX_CACHE=8 GEN_PROMPT='List three primary colors, one per line.' GEN_MAXTOK=16 \
  ./run-chain.sh glm5-chat sparse-moe CPU 8180 9280 \
  'C:\cascadia\models\glm5' node1,node2,node3,node4
```

✅ Expect a coherent instruct answer. First token is slow (NVMe-bound prefill); a warm prefix-cache hit (two-turn `GEN_U1/GEN_A1/GEN_U2` mode) cuts TTFT ~2.15×.

**B4 · streaming check** `[orchestrator]` — launches the chain and streams one completion, timestamping each SSE frame as it arrives (time-separated frames prove per-token streaming):

```bash
GEN_MAXTOK=10 GEN_PROMPT='List three primary colors, one per line.' \
  ./glm5-stream-check.sh gstream 8181 9281
```

✅ Expect `content-type: text/event-stream`, then per-token `delta.content` frames spread out in time (~seconds apart on the I/O-bound path), a final `finish_reason` frame, and `data: [DONE]`.

**B5 · KV-prefix cache warm hit** `[recommended flag]` — two requests sharing a long prefix on the same chain: REQ1 (cold) populates the per-rank cache, REQ2 (warm) hits it and prefills only the suffix — the mechanism behind the recommended `CASCADIA_GLM5_PREFIX_CACHE=8`:

```bash
./glm5-prefix-cache-check.sh gpfx 8184 9284
```

✅ Expect REQ2 much faster than REQ1 and a `kv-prefix cache HIT` log line (the ~2.15× multi-turn TTFT win).

**B6 · long context + bf16-KV** `[recommended flag @ long context]` — needle-in-haystack recall at a few-thousand-token context, run with f32 KV then bf16 KV to confirm bf16 preserves recall (the prerequisite for the recommended `CASCADIA_GLM5_BF16_KV=1`):

```bash
./glm5-longctx-eval.sh glong 8185 9285                                                       # f32 KV
WORKER_EXTRA="$env:CASCADIA_GLM5_BF16_KV='1'; " ./glm5-longctx-eval.sh glongbf16 8186 9286    # bf16 KV
./glm5-kv-quality-eval.sh gkv 8187 9287                                                       # deeper 2-needle parity
```

✅ Expect the planted needle recalled in **both** the f32 and bf16 runs (bf16 KV preserves long-context recall), and the chunked-prefill path handling the >256-token prompt with no `batch_count` error.

**B7 · iGPU/NPU expert offload** `[optional; scale-out only]` — only meaningful once a rank's experts are RAM-resident (shard wide enough). Build the per-expert OV IRs on the export, ship the `experts_ov/` overlay alongside the model, then run with the offload env forwarded to the workers:

```bash
python tools/glm5_expert_ov.py --src glm5-export --validate   # sanity-check int4 vs fp16 vs numpy ref (needs nncf)
python tools/glm5_expert_ov.py --src glm5-export             # convert all MoE layers, --dtype int4 (default)
# copy glm5-export/experts_ov/ next to C:\cascadia\models\glm5 on each node, then:
CASCADIA_GLM5_OV_EXPERTS=1 CASCADIA_GLM5_OV_DEVICE=GPU \
  ./run-chain.sh glm5-ov sparse-moe CPU 8182 9282 \
  'C:\cascadia\models\glm5' node1,node2,node3,node4
```

✅ Only wins when RAM-resident; on a disk-streaming shard it's a no-op-at-best. The IR export + on-device run are unvalidated in CI — this step is the on-hardware validation.

**B8 · async lookahead prefetch + autopin** `[optional; opt-in feature]` — the A/B/C harness compares baseline vs `R1READ` vs `LOOKAHEAD`, per-token-normalized (API `completion_tokens`) with a compute-vs-network-gap split; a warm-up pass plus both run-orders control the warmth / token-count confounds:

```bash
# glm5-abc-check.sh consumes base_api..base_api+3 (warm-up + 3 configs), so give each run a clear range:
GEN_MAXTOK=30 ./glm5-abc-check.sh abc 8193 9293               # base -> r1read -> lookahead (api 8193..8196)
REVERSE=1 GEN_MAXTOK=30 ./glm5-abc-check.sh abcrev 8197 9297   # lookahead cold-first (api 8197..8200)
```

✅ Expect `LOOKAHEAD` to cut per-rank decode **compute ~9%** at flat expert-cache hit % (latency-hiding, output-identical, recall 80–91%), and `R1READ` to lose (page-cache eviction → hit % ~50%). The per-rank worker logs also confirm the **Windows autopin fix** — `mlock'd N/N hot experts` where the pre-fix build showed `0/N`. End-to-end tok/s is dominated by the inter-node network gap, so read the per-rank compute column, not the wall clock.

## Notes

- Tool-calling is engine-agnostic (API-layer parse of the accumulated text buffer, which the streamed deltas reconstruct), so it works for glm5 in both streaming and non-streaming without engine changes.
- See [`docs/architectures/glm5.md`](docs/architectures/glm5.md) for the architecture and design detail.

## Checklist

- [ ] `just check` passes locally (`cargo fmt --check`, `cargo clippy --workspace`, `cargo test --workspace`)
- [ ] Commits follow [Conventional Commits](https://www.conventionalcommits.org/), one logical change per commit
- [ ] Touches the sparse-moe engine — the Rust mmap int4 expert path is the default; the OpenVINO int4 expert backend is opt-in


7i7o (approved):
tACK 👍

Issue #24, opened 2026-05-18
RFC: continuous batching across pipeline-parallel stages


t8:
## Summary (RFC)

**This is a design RFC, not an implementation issue.** Propose the design for cross-host continuous batching across pipeline-parallel stages — a feature that no existing project ships as a first-class capability. If Tahoma implements it, that's a distinctive published artifact.

Goals: collect requirements, scope the design space, identify the key invariants, propose a v0 implementation roadmap before any code lands.

## Why

OVMS continuous-batches within one host. vLLM with Ray supports pipeline-parallel but does **not** continuous-batch cleanly across PP stages — its scheduler is per-replica, not per-stage. `intel/llm-scaler` scales within one chassis only. exo doesn't continuous-batch at all. **No project does cross-host CB+PP.**

Tahoma is uniquely positioned: its multi-host PP transport is already there. Adding CB on top would put Tahoma alone in this niche and is the strongest possible "moat" feature.

But it's hard. The naive "just continuous-batch per stage" doesn't work because:
- Different requests are at different stages at any moment
- Paged-attention KV cache must be coordinated across stages (each stage owns the KV for its layer range)
- Request scheduling decisions on rank 0 must propagate to ranks 1..N-1 with bounded lag
- Cancellation must be coherent across stages

The RFC should answer these questions before any implementation.

### Q1. Scheduler ownership

**Where does the scheduler live?**

- **Option A**: Centralized on rank 0. Rank 0 decides which requests are active in the batch each step; broadcasts decisions to downstream ranks.
- **Option B**: Decentralized. Each rank independently advances whatever the upstream sent. Less coordination but harder to enforce per-step batch composition.
- **Option C**: Hierarchical. Rank 0 owns request admission; downstream ranks just process whatever frames arrive.

**Recommendation prior:** Option A. Simpler reasoning model. The cost is rank 0's decisions add a control-channel hop to every step.

### Q2. Batch composition

If rank 0 has B simultaneous requests, what does each PP stage see per step?

- **Option A (per-step batch)**: All B requests are in flight at every stage. Each step processes 1 token per request × B requests, in parallel through the pipeline. Looks like vLLM's static batching at the PP layer.
- **Option B (interleaved)**: Different requests at different stages. R1 is at layer 30, R2 is at layer 15, etc. Maximizes pipeline utilization but requires per-stage KV cache tagged by `(request_id, stage_id)`.
[...]

PR #116, opened 2026-07-28
feat(ov-genai): continuous batching via ContinuousBatchingPipeline (#20)

### LINKED ISSUE #20: Continuous batching for ov-genai via ContinuousBatchingPipeline

t8:
Closes #20 (single-host scope; cross-stage CB stays RFC #24).

## What

`--cb` on the ov-genai engine swaps `LLMPipeline` for ov-genai's
`ContinuousBatchingPipeline`: concurrent requests join one paged-attention batch
instead of serializing one whole generation per `step()`.

## Pieces

- **Shim**: `cascadia_cb_*` C ABI — pipeline create with `SchedulerConfig`
  (cache_size, max_num_seqs, max_num_batched_tokens, dynamic_split_fuse,
  enable_prefix_caching), `add_request`, `step`, per-handle incremental `read`,
  `cancel`, `count_tokens`. `CbPipeline`/`CbHandle` mirror the existing
  stub/native dual build.
- **Engine**: `OvGenaiCbEngine`. `submit()` enrolls into the running batch
  immediately; `step()` advances the batch one scheduler iteration and emits
  per-task text deltas. **New capability: `cancel()` aborts one request
  mid-generation** — impossible on the monolithic path, where cancel only drops
  queued tasks.
- **CLI**: `--cb` + five `--cb-*` tuning flags, validated (ov-genai only, tuning
  flags require `--cb`, incompatible with `--draft-model`/`--prompt-lookup`/
  VLM-layout exports, rejected on NPU).
- Also fixes per-request `seed` reproducibility on this path.

## Performance — read before enabling

`--cb` is a throughput/latency **trade**, not a strict upgrade. Measured on a
Lunar Lake box (Arc 140V iGPU + CPU) across Phi-3.5-mini-int4 and Qwen3-8B-int4,
short and ~1200-token prompts, concurrency 1–16 (full table in
`docs/engines/ov-genai.md`):

| workload | device | without `--cb` | with `--cb` |
|---|---|---|---|
| short prompt, 16 concurrent | GPU | 36.2 | 176.3 (**4.9×**) |
| short prompt, 8 concurrent | GPU | 39.8 | 141.8 (3.6×) |
| short prompt, 8 concurrent | CPU | 22.7 | 36.8 (1.6×) |
| short prompt, 1 request | GPU | 39.9 | 36.9 (**0.92×**) |
| ~1200-tok prompt, 8 concurrent | GPU | 33.4 | 20.7 (**0.6×**) |
| ~1200-tok prompt, 4 concurrent | CPU | 15.9 | 3.3 (**0.2×**) |

- The 3–5× win is real, and needs short prompts plus ≥4 in-flight requests.
- Every configuration measured loses 8–15% at concurrency 1.
- **On CPU with long prompts throughput collapses ~5×**, on both models, and is
  *not* recoverable by raising `--cb-max-batched-tokens` (which does help on
  GPU). RAG-style traffic on a CPU worker is the worst case for this flag.

The docs say all of this, `--cb` stays opt-in and off by default, and a CPU
worker with `--cb` warns at startup — the collapse has no runtime signal of its
own, so an operator would otherwise just watch it get slower.

## Device semantics

| Device | `--cb` |
|---|---|
| GPU | continuous batching — where the win is |
| CPU | works; modest gain on short prompts, severe loss on long ones |
| NPU | not supported upstream (static NPUW pipeline); rejected by the CLI |

## Testing

Stub-mode (macOS/Linux CI): 54 test binaries green, including 11 new CB engine
state-machine tests behind a `CbPipe` seam (liveness chunk, idle bound at the
real `MAX_IDLE_STEPS`, eviction arm, finish-reason preference and fallback,
per-chunk token accounting, step-failure fan-out, submit poisoning, dedupe/cap,
cancel), 13 pure-Rust UTF-8 hold-back/resync tests, and a runner ordering test.
Each was mutation-checked — re-introducing the bug it names makes it fail.
fmt + clippy clean on changed code. Same suite green on Windows.

On hardware (OV GenAI 2026.2, Arc 140V + CPU): long-prompt smoke, six-identical-
concurrent-prompt determinism, SSE cancel, chat-template parity vs
`LLMPipeline`, and the throughput matrix above.

**Not verified on hardware**: the KV-eviction (`IGNORED`) arm — 8 concurrent
4011-token prompts against a 1 GB cache never triggered it, because OV preempts
rather than dropping and the API rejects over-long single prompts first; the
detokenizer re-anchor branch (unit-tested, never observed firing); and
`MAX_IDLE_STEPS` end-to-end at its real value (verified at a lowered constant,
and unit-tested at the real one).

Cancel is verified on the **streaming** path. On the non-streaming endpoint a
client abort does not cancel — the request runs to completion holding a batch
slot, because hyper only notices the peer is gone when it writes.

## Follow-ups (out of scope)

- API-layer admission (`DEFAULT_MAX_CONCURRENT = 16`) still caps concurrent
  clients; making it configurable is an API change tracked separately.
- Speculative decode inside CB (upstream supports draft models in CBP).
- Cross-stage CB+PP: RFC #24.
- Root-cause the CPU long-prompt collapse; it is not the prefill window.


t8:
## On-hardware e2e: PASS (pawan-01, Lunar Lake Arc 140V iGPU, OV GenAI 2026.2, Qwen3-1.7B-int4-ov)

Ran isolated from the box's live cascadia-node (fresh build dir, ports 18080/19180, path-matched process cleanup; live node verified untouched before/after).

**Two real bugs found and fixed on-HW** (pushed to this branch):
1. `GenerationHandle::drop()` no longer exists in OV GenAI 2026.2 → `cancel()` (frees KV blocks; right semantics for client-disconnect).
2. `completion_tokens` reported `2N-1`: the API sums per-chunk `n_tokens`, so the final chunk's cumulative total double-counted interim deltas. Chunks now carry per-chunk increments; verified exact (`max_tokens=8→8, 16→16`).

**Results:**
- **Delta semantics confirmed**: `max_tokens=24` → exactly 24 single-token deltas, no duplication — `read()` is delta-per-read; the cumulative-text risk flagged in the PR body is refuted.
- **Concurrency (the point of #20)**: 6 concurrent streaming requests — first tokens 1.02–1.48 s, all complete together at 3.19 s (one shared batch). Baseline without `--cb`: silence until **10.55 s**, then every request's full text at once. **~3.3× wall-clock at concurrency 6**, and streaming actually streams (the monolithic path emits one final chunk).
- **Seed reproducibility (acceptance criterion)**: temp 0.8 + seed 42 twice → byte-identical; seed 43 → different. PASS — impossible on the StatefulLLMPipeline path.
- **Cancel**: client disconnect mid-stream; subsequent requests unaffected. PASS.
- Cold CBP compile + warmup on iGPU: ~5 s for 1.7B int4 (fresh CACHE_DIR).

Remaining before merge: nothing blocking from my side — CPU-device run and a longer soak would be nice-to-haves.

7i7o (approved):
tACK 👍

PR #117, opened 2026-07-28
feat(npu): continuous batching on the NPU via packed multi-slot decode

t8:
Continuous batching on the one device where cascadia cannot borrow it from OpenVINO. Paged attention lives in the CPU/GPU plugins; the NPU servable is stateful and sequential. This implements it natively.

**Status: feature-complete and validated on hardware** — exporter, host primitives, engine execution, multi-stage pipeline, per-slot cancel, and prefix caching. Every number below was measured on Lunar Lake (Core Ultra 7 258V / Intel AI Boost), OpenVINO 2026.2.1.

## The finding it rests on

Reshaping a static export's **batch** dimension is rejected by the NPU compiler — identically on both graph shapes tested, at N=2/4/8, while batch-1 compiles fine:

```
Failed Pass ConvertBatchedLayerTo1N
failed to legalize operation 'IE.Convolution' that was explicitly marked illegal
```

The pass *name* is the useful part: the compiler's only strategy for batch>1 is unrolling into N batch-1 ops, so even a legalized batch axis would amortize nothing. But the same export compiles at **seq>1** (chunked prefill has relied on that since #107), and decode is weight-bound, not KV-bound — **458 MB of weights against 29.3 MB of KV per token, 16:1**. The sequence axis is where the amortization is.

So: pack N requests into the sequence dimension, isolate them with a block-diagonal mask. Batch stays 1; the rejected pass never runs.

| seq | ms/iter | ms/token | vs seq=1 |
|---|---|---|---|
| 1 | 10.57 | 10.57 | 1.00x |
| 4 | 25.58 | 6.40 | 1.65x |
| 8 | 27.17 | 3.40 | 3.11x |
| 16 | 28.10 | 1.76 | **6.01x** |

Marginal cost ~0.21 ms per added slot. Isolation is **bit-exact** (`max|delta| = 0` with every other slot randomized); a packed slot equals the same sequence run alone to `3.05e-05` on magnitude ~5806, cosine `1.00000000`.

## What unlocked it

The stock export cannot express per-row masking — it takes a 2D `attention_mask [1, T]` shared by every query row. But it builds its 4D additive mask in a **single node** feeding all SDPA mask inputs, and SDPA's mask slot already carries a query dimension. Replacing that one node with a `[1,1,S,T]` Parameter hands mask construction to the host, where per-slot policy belongs.

## Contents

- **`tools/packed_variant.py`** — performs that graph edit. Runs on an already-exported stage: no HF model, no torch, no re-trace. Also wired as `export_shards.py --packed-slots N`.
- **`packed.rs`** — host-side `PackedKv`: per-slot KV regions, a shared prefix region, and **one** mask writer covering block-diagonal decode, causal prefill chunks, and prefix reuse (mixing them is dynamic-split-fuse for free). Idle rows open on their own column — never fully blocked, which would NaN softmax and poison the live rows.
- **`packed_exec.rs`** — slot table, admission with prefix matching, prefill-first scheduling.
- **`runtime.rs`** — `--packed-slots` / `--packed-prefix`, the multi-stage plan/token wire, per-slot cancel.
- **`docs/perf/NPU_PACKED_SLOTS.md`** — full measurements and sizing.

## End-to-end

`--packed-slots N` turns a static export into a slot-served worker. Validated on Llama-3.2-1B single-stage, **both** devices that run the stateless static path:

| device | | first tokens | all complete | overlap |
|---|---|---|---|---|
| NPU | packed (4) | 2.67–2.85 s | 8.68 s | yes |
| NPU | baseline | 0.32–7.48 s | 11.17 s | no |
| CPU | packed (4) | 2.12–2.93 s | 9.32 s | yes |
| CPU | baseline | 0.57–12.53 s | 18.34 s | no |

Baseline serializes; packed admits all four at once and early finishers retire while others run. Worst-case TTFT 2.6x (NPU) / 4.3x (CPU). Identical text on both devices, and identical to the non-packed path — an end-to-end parity check.

**These 4-slot end-to-end ratios measure scheduling and are bounded by it.** The 3.1x/6.0x above is the graph-level amortization ceiling at 8/16 slots; reaching it end-to-end means exporting more slots and a wider `--static-context`.

## Multi-stage

Stage 0 ships an I64 `[1,3,S]` plan frame — slot id per row (`-1` idle), absolute position, prefix-reuse length — ahead of the `[1,S,hidden]` block; the tail replies one token per row. Relays decode it, re-derive same-slot causal order, and run their own rings. A row whose position equals its reuse length starts that slot's sequence — an in-band signal, so no separate admission message. Every stage must be launched with the same `--packed-slots` (baked into the IR shape).

Validated on 2-stage TinyLlama over loopback: correct output, and 4 concurrent with first tokens 2.09–2.26 s, slot 2 retiring at 5.9 s while others ran to 7.5 s.

## Prefix caching (`--packed-prefix N`)

Paged attention normally provides this, and the NPU has none — but sharing doesn't need paging, it needs a mask that can open the same columns for several rows, which this design already has. The first `N` columns become a read-only shared prefix every slot may attend to. The first request populates it (K/V + token ids); later requests match by longest common prefix, start at absolute position `k`, and never re-prefill those tokens. RoPE stays correct because the cached K/V were computed at their true positions.

NPU, 4 slots, four requests sharing a ~96-token system prompt:

| | first request | later requests | speedup |
|---|---|---|---|
| `--packed-prefix 96` | 2.26 s | **0.38 s** | **5.95x** |
| off | 2.81 s | 2.09 s | 1.34x (warmup only) |

Answers unchanged (Paris / Tokyo / Rome / Madrid) — the check that reuse doesn't corrupt attention.

## Per-slot cancel

A disconnecting client's slot retires and its KV region returns to the free pool immediately, leaving other in-flight slots untouched. The single-task path can only drop *queued* tasks. Observed live: four disconnects released slots one at a time (`in_flight=3,2,1,0`).

## Accuracy: no measured accuracy loss

Harness committed at [`tests-e2e/accuracy`](../tree/feat/npu-packed-slots-cb/tests-e2e/accuracy). Every comparison is a **full-text exact** match — never a substring or first-sentence check, which would pass while later output degraded.

**Scored against ground truth** (this is the accuracy claim; the parity suites below only establish equality against a reference, never whether an answer is *right*):

| benchmark | baseline | packed (4 slots) | delta |
|---|---|---|---|
| short-answer, 20 factual Q&A | 15/20 (75%) | **15/20 (75%)** | **+0** |
| long-form, 7 tasks, ordered-atom recall | 89.5% | **89.5%** | **+0.0 pts** |

Long-form is scored on **ordered atom recall** — every expected item must appear in sequence — so omission, reordering, truncation and repetition loops all cost score. Repetition ratio identical (0.081 both). Text identical on 20/20 and 7/7 respectively. The five short-answer failures are the 1B model's own and identically wrong in both configs.

**Safety properties, NPU single-stage e2e** (Llama-3.2-1B, 128-token generations):

| property | result |
|---|---|
| determinism (solo vs solo-repeat) | **10/10 exact**, both configs |
| **batch-composition invariance** — output must not depend on batch-mates | **10/10 exact**, both configs |
| prefix cache on vs off (sequential / concurrent / internal) | **8/8 / 8/8 / 8/8 exact** |
| packed vs baseline, free-form | 8/10 exact; 92.8% character agreement |

**Graph-level** (Qwen2.5-1.5B, NPU): row isolation `max|delta| = 0` at N=4/8/16 with every other slot randomized; a packed slot vs the same sequence run **alone** differs by `3.05e-05` on magnitude ~5806, **cosine 1.00000000**. Teeth check moves row 0 by ~1.6e3, so isolation is not vacuous.

### Where text is *not* identical, and why

Two of ten free-form 128-token prompts diverge: `about`/`approximately` at 72.5% depth, and `1,143`/`1,145` at 66.3%. **One token each, re-converging immediately**, identical word counts, no repetition or truncation. That is an fp16 near-tie flipping `argmax` — packed and baseline are two different compiled graphs.

Multi-stage (TinyLlama 2-stage) is 5/10 against baseline, and a control settles that it is shape sensitivity rather than a wire fault:

| comparison | exact |
|---|---|
| baseline vs packed slots=4 | 5/10 |
| **packed slots=4 vs packed slots=2** | **3/10** |

Two configs of the *same code*, differing only in slot count, agree **less** than packed agrees with baseline — because changing `packed_slots` changes the compiled graph shape and the prefill chunk width. Same sensitivity `chunk_take` already documents for `--static-prefill-seq` ("forks observed as early as token 2"). Determinism and batch-composition invariance are 10/10 in every multi-stage config.

**Fair summary:** no measured accuracy loss, no corruption, no cross-request leakage, no degeneration. Output is **not** guaranteed bit-identical to the unpacked path on long free-form generation — inherent to changing the compiled graph under greedy decode, not specific to this implementation.

### What the accuracy work found

Four defects, none visible to the functional e2e tests:

1. **`completion_tokens` reported 2N−1** (cap 8 → 15, cap 16 → 31) — the final chunk carried the cumulative total while the API sums per-chunk counts. Same defect I fixed on the ov-genai cb path and failed to carry over. Fixed; verified 8→8, 16→16 with byte-identical text.
2. **Wire dtype mismatch** — packed sent hidden states as f32 while the non-packed path sends f16, doubling bytes per stage hop for nothing. Fixed.
3. **Runner chunk reordering** — the first run failed batch-invariance on *both* configs with scrambled word order; that was the pre-existing bug fixed on main by `abccce5`, which my test binary predated. Not a packing fault, but the suite is sensitive enough to catch that class.
4. **EOS token counting** — packed counts the EOS token in `completion_tokens`, the non-packed path does not, so EOS-terminated completions read +1 against baseline (length-capped match exactly). Packed's count is the accurate one; documented rather than degraded to match.

### Gaps I am not papering over

- Scored long-form tops out at **110 generated tokens** — the list tasks that make scoring possible terminate early on EOS. The longest generations I have (128 tokens, free-form) are exactly where the single-token divergences appear, and are unscoreable.
- Multi-stage exact text equality against baseline is **not established** and, per the control above, is not available across differing shapes.
- The 2-stage pipeline **hangs under sustained concurrent load in both packed and baseline configs** — a separate pre-existing issue found here, not diagnosed.

## Honest limits

- **Fixed partition, not paging.** `region = (static_context - 1 - prefix) / slots`, contiguous and uniform. You cannot give one request 900 tokens and three requests 40; a lightly-used slot still reserves its full region. Slot count is baked into the IR shape, so changing it means rebuilding the variant and restarting every stage.
- **Prefix cache is one entry**, populated by whichever request arrives first, no LRU, no block-level dedup between partially-overlapping prompts, no eviction under pressure.
- **Sizing is a real trade-off**: slots and prefix both come out of the same fixed window. High slot counts with useful contexts need a wider `--static-context` at export.
- Greedy sampling only on this path (inherited from the static path).
- **`--packed-slots` means something different per device in cascadia**: on CPU/GPU the ov-genai path (#116) gets OV's paged attention and its prefix caching; this is the NPU-native complement.

## Testing

18 packed unit tests (mask layout incl. shared-prefix visibility, cross-slot blocking, idle-row safety, causal chunks, region-local scatter, in-region slide, prefix populate/reuse/position-shift, plan + token frame round-trips and bad-frame rejection, relay-rebuilt-plan mask equality), 85 engine tests and 78 CLI tests total, plus the cross-feature gate pinning `--cb` and `--packed-slots` mutually exclusive.

Hardware: graph-level isolation/equivalence on NPU; single-stage e2e on NPU **and** CPU; 2-stage pipeline e2e; prefix-cache e2e; per-slot cancel observed live; and the three accuracy suites above.

Related: #20 / #24 cover continuous batching on the ov-genai CPU/GPU path.

### DIFF
--- a/crates/cascadia-api/src/lib.rs
+++ b/crates/cascadia-api/src/lib.rs
@@ -1786,6 +1786,9 @@ fn engine_error_response(err: cascadia_engine::EngineError) -> axum::response::R
     let status = match &err {
         EngineError::QueueFull { .. } => StatusCode::SERVICE_UNAVAILABLE,
         EngineError::NotLoaded | EngineError::NotConnected => StatusCode::SERVICE_UNAVAILABLE,
+        // Permanent for this request: retrying sends the identical prompt into
+        // the identical window. 413 is what every SDK treats as non-retryable.
+        EngineError::PromptTooLong(_) => StatusCode::PAYLOAD_TOO_LARGE,
         EngineError::InvalidConfig(_)
         | EngineError::PeerRejected(_)
         | EngineError::ShardRejected(_) => StatusCode::BAD_REQUEST,
@@ -1821,6 +1824,19 @@ mod tests {
     // engine sets `n_tokens` on a single text-bearing final chunk, that count
     // is authoritative (not the "1 per chunk" fallback), so `usage` reflects
     // the real generated-token count rather than 0/1.
+    /// A prompt that cannot fit the engine's per-request window is a permanent,
+    /// deterministic rejection of THIS request — not backpressure. It must not
+    /// map to 503: the official SDKs retry 5xx with backoff, so a request that
+    /// can never succeed would be sent three times before surfacing, and the
+    /// operator could not tell it apart from genuine "engine at capacity".
+    #[test]
+    fn prompt_too_long_maps_to_413_not_503() {
+        let r = engine_error_response(cascadia_engine::EngineError::PromptTooLong(
+            "prompt is 207 tokens but this worker's per-slot KV region holds 127".into(),
+        ));
+        assert_eq!(r.status(), StatusCode::PAYLOAD_TOO_LARGE);
+    }
+
     #[test]
     fn chunk_token_count_honors_engine_count() {
         // ov-genai's shape: whole response on one final chunk, n_tokens set.
--- a/crates/cascadia-cli/src/lib.rs
+++ b/crates/cascadia-cli/src/lib.rs
@@ -424,6 +424,14 @@ pub struct WorkerArgs {
     #[arg(long, default_value_t = 0)]
     pub packed_slots: u32,
 
+    /// Reserve N KV slots as a read-only SHARED prefix that every packed slot
+    /// may attend to — prefix caching without paged attention. The first
+    /// admitted request populates it; later requests sharing that prompt prefix
+    /// skip re-prefilling those tokens. Taken from the same window, so it costs
+    /// per-slot context. Requires --packed-slots. 0 = off.
+    #[arg(long, default_value_t = 0)]
+    pub packed_prefix: u32,
+
     /// Continuous batching (#20, ov-genai only): serve concurrent requests
     /// through one ContinuousBatchingPipeline (paged attention; CPU/GPU
     /// plugins) instead of one generation at a time. Incompatible with
@@ -661,6 +669,7 @@ impl WorkerArgs {
             spec_k: 5,
             prompt_lookup: 0,
             packed_slots: 0,
+            packed_prefix: 0,
             cb: false,
             cb_cache_size: 0,
             cb_max_num_seqs: 0,
@@ -1245,6 +1254,7 @@ fn build_builder(args: &WorkerArgs) -> Result<Box<dyn Builder>> {
             }
             b = b.with_chunked_prefill_disabled(args.no_chunked_prefill);
             b.packed_slots = args.packed_slots;
+            b.packed_prefix = args.packed_prefix;
             b = b.with_prefill_parking(args.park_prefill);
             b = b.with_gemv_offload(args.gemv_offload);
             if let Some(dir) = resolve_ov_cache_dir(args.ov_cache_dir.as_deref()) {
@@ -1498,6 +1508,12 @@ fn validate_worker_runtime_flags(args: &WorkerArgs) -> Result<()> {
     if (args.prefill_device.is_some() || args.park_prefill) && args.no_chunked_prefill {
         return Err(anyhow!(
             "--prefill-device / --park-prefill conflict with --no-chunked-prefill"
+        ));
+    }
+    if args.packed_prefix > 0 && args.packed_slots == 0 {
+        return Err(anyhow!(
+            "--packed-prefix requires --packed-slots (the shared prefix lives in the packed \
+             KV window)"
         ));
     }
     // Packed multi-slot decode lives in the ov-runtime static (NPU-target) path.
@@ -2389,6 +2405,17 @@ mod python_tests {
 
         let mut a = worker("m", EngineKind::OvRuntime);
         a.packed_slots = 8;
+        assert!(validate_worker_runtime_flags(&a).is_ok());
+
+        // the shared prefix lives inside the packed window, so it needs slots
+        let mut a = worker("m", EngineKind::OvRuntime);
+        a.packed_prefix = 128;
+        let err = validate_worker_runtime_flags(&a).unwrap_err().to_string();
+        assert!(err.contains("--packed-slots"), "{err}");
+
+        let mut a = worker("m", EngineKind::OvRuntime);
+        a.packed_slots = 4;
+        a.packed_prefix = 128;
         assert!(validate_worker_runtime_flags(&a).is_ok());
     }
 
--- a/crates/cascadia-cli/src/lib.rs
+++ b/crates/cascadia-cli/src/lib.rs
@@ -417,6 +417,13 @@ pub struct WorkerArgs {
     #[arg(long, default_value_t = 0)]
     pub prompt_lookup: u32,
 
+    /// Packed multi-slot continuous batching (NPU, ov-runtime static exports):
+    /// serve N concurrent requests in ONE inference by packing them into the
+    /// sequence dimension with a per-row mask. Requires a packed variant beside
+    /// the decode IR (`tools/packed_variant.py --slots N`). 0 = off.
+    #[arg(long, default_value_t = 0)]
+    pub packed_slots: u32,
+
     /// Continuous batching (#20, ov-genai only): serve concurrent requests
     /// through one ContinuousBatchingPipeline (paged attention; CPU/GPU
     /// plugins) instead of one generation at a time. Incompatible with
@@ -653,6 +660,7 @@ impl WorkerArgs {
             draft_device: None,
             spec_k: 5,
             prompt_lookup: 0,
+            packed_slots: 0,
             cb: false,
             cb_cache_size: 0,
             cb_max_num_seqs: 0,
@@ -1236,6 +1244,7 @@ fn build_builder(args: &WorkerArgs) -> Result<Box<dyn Builder>> {
                 b = b.with_prefill_device(dev);
             }
             b = b.with_chunked_prefill_disabled(args.no_chunked_prefill);
+            b.packed_slots = args.packed_slots;
             b = b.with_prefill_parking(args.park_prefill);
             b = b.with_gemv_offload(args.gemv_offload);
             if let Some(dir) = resolve_ov_cache_dir(args.ov_cache_dir.as_deref()) {
@@ -1491,7 +1500,27 @@ fn validate_worker_runtime_flags(args: &WorkerArgs) -> Result<()> {
             "--prefill-device / --park-prefill conflict with --no-chunked-prefill"
         ));
     }
-    // Continuous batching (#20) lives in the ov-genai CBP path only.
+    // Packed multi-slot decode lives in the ov-runtime static (NPU-target) path.
+    if args.packed_slots > 0 {
+        if args.engine != EngineKind::OvRuntime {
+            return Err(anyhow!(
+                "--packed-slots requires --engine ov-runtime (packed multi-slot decode is a \
+                 static-KV path feature)"
+            ));
+        }
+        if args.packed_slots < 2 {
+            return Err(anyhow!("--packed-slots must be 0 (off) or >= 2"));
+        }
+        if args.total != 1 {
+            return Err(anyhow!(
+                "--packed-slots is single-stage only for now (--total 1); the multi-stage wire \
+                 does not yet carry [1, S, hidden] + per-slot positions"
+            ));
+        }
+    }
+    // Continuous batching (#20) lives in the ov-genai CBP path only. It is a
+    // different mechanism to --packed-slots above: OV's paged attention on the
+    // CPU/GPU plugins, versus our sequence-packing on the NPU static path.
     if (args.cb
         || args.cb_cache_size > 0
         || args.cb_max_num_seqs > 0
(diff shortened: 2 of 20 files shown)


t8:
## Engine execution path wired + end-to-end validated

Follow-up commits on this branch take the substrate to a working feature.

**Wired** (`05b67e8`): `PackedState` owns the compiled packed variant, the per-slot KV regions and the slot table. `--packed-slots N` dispatches `step_first` to the packed path — admission tokenizes into a free slot, one inference either prefills a chunk or decodes a row per ready slot, and a finished slot **retires immediately**, freeing its KV region for the next admission rather than waiting for the batch. Sampling is per row via the existing `argmax_logits_row`. Gated to `--engine ov-runtime` + `--total 1`, with CLI tests.

**Two bugs the e2e caught** (`f88e478`), neither visible to unit tests:
1. A 42-token prompt at `packed_seq=4` needs 11 prefill inferences, and intermediate ones emit no chunk — so the runner's "no progress for 3 consecutive steps" guard closed the stream before the first token ever arrived. `step_first_packed` now keeps inferring until it produces a chunk, which is what the single-task static path already does with its prompt.
2. `prompt_tokens` was missing from the packed final chunk, so `/v1` usage read 0. (The non-packed static path still reports 0 there; the packed path is now better.)

**End-to-end** on Llama-3.2-1B-Instruct, single-stage int4 static export. Run on the **CPU** device — which executes the stateless static path exactly as the NPU does — so concurrency semantics are exercised without contending for an NPU:

| | first tokens | all complete | overlap |
|---|---|---|---|
| `--packed-slots 4` | 2.12–2.93 s | **9.32 s** | yes |
| baseline (packing off) | 0.57–12.53 s | **18.34 s** | no |

Baseline serializes — each request's first token lands only after the previous finishes. Packed admits all four at once (log confirms slots 0/1/2/3), and the two that hit EOS early retired at 5.5 s and 5.9 s while the others ran on: continuous, not batch-synchronous. Wall clock ~2.0×; worst-case TTFT 12.53 s → 2.93 s (4.3×).

Both configurations return **identical text** for the same prompt ("The capital of France is Paris."), which is an end-to-end parity check between the packed and single-task paths. Usage now reads `prompt_tokens: 42, completion_tokens: 15`.

These CPU ratios measure *scheduling*, not the weight-amortization win — that's the NPU property (16:1 weight:KV) in the tables above, where packing measured 3.1× at 8 slots and 6.0× at 16.

**Still not wired**: the multi-stage wire carrying `[1, S, hidden]` + per-slot positions, hence the `--total 1` gate. Mid-generation cancel of one slot is also unimplemented (`retire()` exists; nothing calls it from `cancel()` yet).

Test hygiene: ran on pawan-01 against an isolated copy of the model (shared model dir untouched), CPU-only, on ports 18090/19190. A colleague's node on that box was never touched — it went down mid-session because they stopped it to rebuild (cargo/rustc running in their own workspace), and my stop script matches only this branch's exe path.

t8:
## Multi-stage wire + per-slot cancel — both wired and validated

### Multi-stage (`--total 1` gate lifted)

Stage 0 ships an I64 `[1, 2, S]` **plan frame** — slot id per row (`-1` = idle), absolute position per row — ahead of the `[1, S, hidden]` block; the tail replies with one token per row. Relay and head stages decode the plan, re-derive same-slot causal order, and run their own packed inference over their own per-slot rings.

The design choice worth reviewing: a row at **position 0 resets its slot**. That's the same in-band new-sequence signal the single-task static path already uses, so a downstream stage needs no separate admission or eviction message — the plan frame alone keeps every stage's rings in lockstep. Codecs reject dtype/rank/length mismatches, out-of-range slot ids and negative positions rather than indexing or wrapping downstream (a scalar position frame from a non-packed peer is a clean error). Every stage must be launched with the same `--packed-slots`, since the slot count is baked into each stage's IR shape.

**Validated** on a 2-stage TinyLlama static export, both ranks on one box over loopback, CPU device:

- sanity: `"The capital of France is Paris."`, usage `prompt 35 + completion 15`
- 4 concurrent across the pipeline: first tokens **2.09–2.26 s**, slot 2 retired at **5.9 s** while the others ran to 7.5 s, `OVERLAP: True`

### Per-slot cancel

A disconnecting client's slot is retired and its KV region returned to the free pool immediately, leaving the other in-flight slots sharing the inference untouched. The single-task path can only drop *queued* tasks; it can never reclaim a running one. Observed live in the 2-stage run — four disconnects released slots one at a time (`in_flight=3, 2, 1, 0`).

### One bug the e2e caught

`exchange_packed_downstream` / `step_relay_packed` called `Handle::block_on` directly, which panics with *"Cannot start a runtime from within a runtime"* the moment the runner polls the chunk stream on an async worker thread. Fixed to use the engine's existing `self.block_on` helper (`block_in_place` when on an async worker, naked `block_on` otherwise) — the same dispatch the non-packed wire path already uses. Unit tests could not have caught this; it only appears with a real two-process pipeline.

67 engine tests green, including plan/token frame round-trips, bad-frame rejection, and a check that a relay-rebuilt plan masks **identically** to what the sender masked.

Remaining known gap: per-slot contexts are `(static_context - 1)/slots`, so a deployment wanting large contexts at high slot counts must export with a bigger `--static-context`. Sizing is documented.

t8:
## Confirmed on the NPU device

Ran the full packed path on **Intel AI Boost** (pawan-01), Llama-3.2-1B-Instruct single-stage static export, 4 slots. Same binary and model as the earlier CPU run — only the compile target changed.

```
sanity   "The capital of France is Paris."   usage 42 + 15   (identical to CPU)

              first tokens        all complete   overlap
packed (4)    2.67 - 2.85 s        8.68 s         yes
baseline      0.32 - 7.48 s       11.17 s         no
```

Baseline serializes — request 1's first token doesn't arrive until 7.48 s, after two others have finished. Packed admits all four at once and the early finishers retire while the rest run on. Worst-case TTFT **7.48 s → 2.85 s (2.6×)**; wall clock 1.3×.

For the CPU comparison on the same model: packed 9.32 s vs baseline 18.34 s, worst-case TTFT 12.53 s → 2.93 s. Both devices return byte-identical text, which is what you'd expect — it is the same stateless static IR, only compiled elsewhere.

**Reading these numbers honestly:** the end-to-end ratios at 4 slots measure *scheduling*, and are bounded by it. The weight-amortization ceiling is the graph-level property measured earlier (3.1× at 8 slots, 6.0× at 16); reaching it means exporting with more slots and a wider `--static-context`, not more concurrency at N=4. I'd rather show the modest honest number here than quote the 6× as if it were end-to-end.

### One optimization the NPU run motivated

Packed mode was still compiling the chunked-prefill variant, which it never uses — a packed plan whose rows all belong to one slot *is* a causal chunk. Now skipped, saving a full NPU compile (~100 s) and a second resident weight copy. Visible in the load log:

```
packed multi-slot variant: slots=4 seq=4 context=1027 per-request context=255
packed slots: skipping the chunked-prefill variant (packed inference covers prefill natively)
```

Test hygiene: the box's other occupant runs on **GPU** (checked their `node.toml` before touching the NPU), so there was no device contention; their node is up and untouched, box back to 24.8 GB free with no leftovers.

t8:
## Rebased onto main (conflicts resolved) — still open, not merged

`main` gained #116 (ov-genai continuous batching) plus a release, which collided with this branch in `crates/cascadia-cli/src/lib.rs`. Merged `origin/main` in rather than rebasing, since the branch is pushed and earlier PR comments reference its commits.

All three conflicts were **additive** — both sides added flags, initializers and validation at the same spot — so both survive. They are independent features on different engines:

- `--cb` / `--cb-*` → ov-genai, OV's paged attention, CPU/GPU only
- `--packed-slots` / `--packed-prefix` → ov-runtime, sequence-packing, NPU-native

While resolving I also:

- **Fixed a stale comment of mine**: the multi-stage plan frame is `[1,3,S]` (slot, position, prefix-reuse length), not `[1,2,S]` — the third row arrived with prefix caching.
- **Cross-referenced the two mechanisms** in both gates, and pointed main's NPU rejection message at the alternative: `--cb` on NPU now says *"use --engine ov-runtime with --packed-slots for NPU concurrency"* instead of just refusing.
- **Added a test** pinning that `--cb` and `--packed-slots` can never both apply — whichever engine is named, the other flag's gate fires. That interaction is exactly what a merge like this can silently break.
- **Documented both flags in `docs/CLI.md`**, since main now documents the `--cb` family there.

Verification after the merge: full workspace builds, `cargo test --workspace --lib` all green (78 CLI, 67 engine-openvino, 85 api, 59 sparse-moe, etc.), fmt clean.

`mergeStateStatus: CLEAN`. Leaving it open per your instruction.

7i7o (approved):
tACK 👍

PR #128, opened 2026-08-07
fix: multi-stage packed pipeline wedge — rank 0 starved its own I/O driver (#122)

### LINKED ISSUE #122: Multi-stage packed decode loses a token frame and wedges the pipeline
`--packed-slots` with `--total > 1` is **gated off** in the CLI (PR #117) because the multi-stage wire intermittently drops a token frame and rank 0 then blocks on a reply that never arrives. The packing itself is correct — single-stage packed is verified across 5 models / 3 families. This tracks the transport fault.

## Symptom

The pipeline **stops serving** (not: produces wrong text). Once wedged it never recovers.

One exchange is: rank 0 sends a plan frame, sends a hidden frame, awaits a token frame; rank 1 receives the pair, runs its inference, replies. Lockstep, one reply per pair.

With a sequence counter threaded through both sides, at the wedge:

```
rank 1   plan decoded, awaiting hidden    seq=1119
rank 1   hidden received                  seq=1119
rank 1   replying with tokens             seq=1119   <- sent
rank 1   lock acquired, awaiting plan     seq=1120   <- moved on

rank 0   sending plan                     seq=1119  active=[0,0,0,0]
rank 0   both frames sent, awaiting tokens seq=1119
rank 0   (nothing further)
```

Line counts agree independently: rank 1 completed **1120** cycles, rank 0 completed **1118** and is stuck on **1119**. **No WARN or ERROR on either side.** Rank 1's `block_on` returned and it advanced 162 µs later, so its send completed at the API level.

**Rank 1 wrote the frame; rank 0's `recv()` never woke.**

## Conditions

- Needs sustained load: a fresh pipeline serves 4 concurrent 128-token requests fine (12.1 s). It wedges on the first concurrent batch after ~10–20 solo requests.
- Intermittent, roughly two runs in three. Adding instrumentation masked it once — treat as a race; observation perturbs it.

## Ruled out

- **The packing / masking / KV rings.** Single-stage packed verified on Llama-3.2-1B/3B, Llama-3.1-8B, Qwen2.5-14B, Phi-3.5-mini, incl. 8 concurrent identical prompts byte-identical.
- **Slot-state desync between ranks.** First hypothesis; the logs disprove it — rank 1 processed the frame correctly and completely, then replied.
- **A swallowed application error.** The relay loop does swallow non-fatal `Err` from `step()` and would deadlock rank 0 — but no such log appears. (Still a *separate* latent deadlock worth fixing: rank 1 should tell rank 0 it failed rather than silently loop.)
- **Frame size.** The wedging exchange was a 4-row prefill chunk, but decode exchanges send identically shaped frames and pass fine.
- **The f16 wire change.** Reproduces with it in.
- **Multi-stage baseline.** `--total 2` without `--packed-slots` runs the accuracy suite clean. It needs packed + multi-stage together.

## Where to look

Below the engine, in `cascadia-transport`. Engine side is `exchange_packed_downstream` / `step_relay_packed` in `crates/cascadia-engine-openvino/src/runtime.rs`; both behaved correctly.

1. **Framing on a shared duplex session.** Rank 0 holds `down.lock().await` across send-send-recv. Rank 1 *releases* the upstream lock between its recv-pair and its send, re-acquiring it to reply — check whether that can interleave with anything else on the session.
2. **A lost wakeup rather than lost bytes.** Rank 0 is parked in `recv().await`; the bytes may still be in the socket buffer. Dump socket state at the wedge before concluding the write was lost.
3. **`block_in_place` dispatch asymmetry.** Both stages bridge sync→async via `cascadia_runner::run_async`, which picks `block_in_place` vs naked `block_on` by thread context. Rank 0 runs inside `ChunkStream::poll_next` on a runtime worker; rank 1 runs in `run_relay_loop` on a `spawn_blocking` thread that *enters the guard*. Different dispatch per side; `af3c91c` already fixed one bug here.

## Reproducing

Export a 2-stage shard **with** a prefill variant so the baseline is a fair control:

```bash
python tools/export_shards.py --model unsloth/Llama-3.2-1B-Instruct \
  --output-dir <dir> --num-stages 2 --target npu \
  --static-context 1024 --static-prefill-seq 64 --packed-slots 4
```

Start rank 1 first, then rank 0, both with the same `--packed-slots 4`. Drive 10–20 solo requests at `max_tokens=128`, then 4 concurrent. The CLI refuses this configuration now — temporarily revert the `--total 1` check in `validate_worker_runtime_flags` to reproduce.

Notes that cost time:
- Log **each rank to its own file on the node**; worker stdout through an ssh pipe buffers and loses exactly the lines that matter.
- The accuracy harness's per-request timeout is 900 s, so a wedge there looks like a very long hang. Use a short timeout when probing deliberately.
- `recv()` on the reply is now bounded by `CASCADIA_ACTIVATION_TIMEOUT_SECS`, so a wedge surfaces as an error rather than an infinite hang — check the log, not just the client.

## Also blocked on this

`packed_accuracy.py`'s multi-stage parity number was never obtained: the capture wedges before finishing. The single-stage control on Llama-3.2-1B is 8/10 with two benign single-token late diffs (reproduced twice).

t8:
# Fix #122: multi-stage packed decode wedge — driver starvation, not a lost frame

Closes #122.

## Root cause

The token frame was never lost and never late — **rank 0's process starved its own tokio I/O + timer driver**, so the frame sat unread in the socket buffer while the `recv()` future's readable event was never dispatched and no recv deadline could fire either (timers ride the same driver). That is exactly the reported signature: rank 1 wrote the reply and moved on, rank 0's `recv()` never woke, no WARN or ERROR on either side, wedged forever.

The starvation is structural, in `cascadia-runner`: the engine sits behind a **sync** `parking_lot::Mutex`, `step()` holds it across OV inference plus a full plan+hidden/token network round trip, and four different async-context callers took that mutex with a **hard blocking `lock()` on tokio worker threads**:

- `ChunkStream::poll_next` (every concurrent stream's SSE task)
- `Runner::submit` via the API's `generate()` call sites (every arriving request)
- `Runner::cancel` (DELETE handler)
- `ChunkStream::drop` (every client disconnect / completed stream)

A blocked worker is invisible to the scheduler — no core handoff. Once (streams polling) + (requests arriving) + (streams dropping) ≥ worker threads while a step is in flight, **every** worker is pinned, nobody can poll the shared I/O + timer driver, and the step-holder's `recv()` — parked via `block_in_place` + `block_on` — can never be woken. The engine mutex then never releases, so the blocked workers never unblock: permanent, silent, and it needs sustained load plus a concurrent burst to line up, which is why a fresh pipeline served 4 concurrent requests fine and the wedge showed up "after ~10–20 solo requests, on the first concurrent batch", roughly two runs in three.

## Reproduction

`crates/cascadia-runner/tests/packed_wire_stress.rs` rebuilds the production shape with no OpenVINO: a driver runtime and a relay runtime (two processes in prod), real loopback TCP through `cascadia-transport`, identical lock structure and `block_in_place`/`block_on` dispatch on each side, and the issue's load pattern (solo requests, then a concurrent batch).

- **Pre-fix**: wedges deterministically on the first concurrent batch once streams+submits reach the worker count — stuck at exchange 129 with zero progress through a 130 s watchdog, no error surfaced (timers starved too).
- **Post-fix**: all three scenarios (prod-like 8×4, small-box 4×6, disconnect burst) pass in ~1.2 s, 10/10 consecutive runs.

Real-engine A/B on miner (Xeon, OV 2026.2, real `cascadia worker` binaries, 2-stage Llama-3.2-1B packed export per the issue's repro command, rank 0 pinned to 4 cores):

- **baseline** (= `origin/main` 1a9c147f + the CLI gate bypassed): 15 solo requests fine (1–4 s each), then **wedged on the first 6-concurrent batch** — all six hung, `PROBE_FAILED: TimeoutError`. Forensics captured live at the wedge:
  - `ss -tinm`: rank 0's data socket showed **`Recv-Q = 52` for minutes** — exactly one I64 `[1,1,4]` token frame (20 B header + 32 B payload, a 4-row prefill chunk — the same frame shape the issue logged as `active=[0,0,0,0]`) delivered, acked, and **never read**. Rank 1's side: `Send-Q = 0`, all bytes acked. The frame was not lost.
  - `/proc/<rank0>/task/*/wchan`: **all 5 `tokio-rt-worker` threads in `futex_do_wait`; not one tokio thread in `ep_poll`** (the only epoll waiter was the unrelated mDNS daemon). Nobody could poll the I/O + timer driver.
  - Bonus: the wedged rank 0 **survived SIGTERM** (signal handling is a task on the starved runtime) and had to be SIGKILLed — a wedged node also resists orderly supervision.
- **fixed** (= this branch, same load, same 4-core rank 0): the first concurrent batch — the exact point the baseline died — completed in 2.9–7.4 s per request, and the full probe ran **`ALL_OK requests=84 wall=174.2s`** (4 rounds of 15 solo + 6 concurrent, zero errors, zero hangs). Server logs across the whole fixed run (probe + all accuracy captures, 134 requests): **zero WARN/ERROR, zero aborts, zero NACKs** — the new recovery paths exist but never needed to fire.

## The fix

**`cascadia-runner` — never hard-block a worker on the engine mutex (the wedge fix):**

- `poll_next`: `try_lock`; on contention, register a waker and return `Pending`. Every engine-lock release (step, submit, cancel, close) wakes the parked streams; registration re-checks the lock so a release can't slip through the gap.
- `generate_async`: submit dispatched via `spawn_blocking`; the API's four call sites use it. (`generate`/`submit` remain for sync callers, documented as blocking.)
- `cancel` and stream `Drop`: `try_lock`, deferring the engine-side cancel to the next lock holder — the same effective timing a blocking cancel had, since a running step can't be interrupted anyway.
- `close()` wakes parked streams so teardown can't strand them.

**`cascadia-engine-openvino` — harden the packed wire (so a genuinely lost/unanswerable frame degrades cleanly, plus the latent deadlock the issue called out):**

- `exchange_packed_downstream`: the owed token reply now uses `recv_reply` / `recv_reply_prefill` (prefill chunks get the widened budget) instead of idle-tolerant `recv()` under an external `tokio::time::timeout`. The deadline lives inside the transport: a miss can't cancel the recv future mid-frame (which silently discarded partially-read bytes), and a failed reply poisons the connection so a late frame can never be read into the next exchange. Poisoned-socket errors are connection-fatal → a middle rank exits for supervisor rebuild.
- `step_relay_packed`: the hidden frame (owed second half of the pair) is a deadlined `recv_reply`, and the plan is decoded only after the hidden frame is consumed, so a bad plan can't desync the stream. Any failure after the pair is consumed **answers the upstream with a NACK** (an empty `[1,1,0]` token frame) before surfacing the error — previously the relay loop swallowed the error and looped while rank 0 blocked forever. Middle stages propagate the NACK.
- `step_first_packed`: a failed shared inference or exchange **aborts the whole in-flight batch** — every active slot retired (KV region freed), every affected task gets its own attributed error chunk. Previously a bare `Err` killed whichever stream happened to be polling and left the other slots wedged-active. Admission failures are attributed to the failing task. Downstream slot state after an abort is stale but harmless — the next admission resets its slot in-band.

**`cascadia-cli`:** the `--packed-slots` + `--total > 1` gate is removed.

## Multi-stage packed accuracy (the blocked deliverable)

`packed_accuracy.py` now runs to completion multi-stage (it wedged before). 2-stage packed vs 2-stage plain on the same shards, Llama-3.2-1B, CPU device, 128 tokens, exact full-text match:

| check | result |
|---|---|
| packed determinism (solo vs solo-repeat) | **10/10** |
| packed batch-composition invariance | **9/10** (one `can't`→`cannot` flip at char 207 of 477) |
| plain determinism | 10/10 |
| packed vs plain parity (solo / batched) | 5/10 / 5/10 |

Every parity diff diverges **late** (char 93–471 into 277–563-char outputs) into semantically equivalent phrasing — the packed graph's known numeric-drift envelope (issue #122's own single-stage control was 8/10 with the same class of benign late single-token diffs; the multi-stage batch-invariance 9/10 here is actually better than that control). This drift is packed-vs-plain computation difference by construction (block-diagonal mask, region-local KV), not a wire/scheduling artifact: within-packed determinism is 10/10, and no error path fired during any capture.

## Testing

- `cargo test` across the workspace: green (one pre-existing failure on pristine `main`, `cascadia-tests-e2e::binary_serves_health_then_models_then_chat` — `/metrics` 404 — unrelated, fails identically without this branch).
- New: 3 stress scenarios (above), NACK/prefill-detection unit tests in the engine, CLI gate test updated.
- fmt + clippy clean on touched code.

## Notes

- E2E substrate: miner (Linux/CPU) rather than the 2-box NPU rig — the defect and fix live in the tokio/runner/transport layers, which are device-independent (the CPU repro reproduced the issue's NPU signature down to the identical 4-row frame size), and the Lunar Lake boxes host live demos. A confirmation run on the NPU rig is a cheap follow-up if wanted: the exact export + commands from the issue now work gate-free.

- Open PR #95 (bounded token recv + seq echo on the **non-packed** head path) attacks an adjacent failure class at the transport layer; this PR deliberately doesn't touch those call sites, and the two compose.
- The `docs/perf/NPU_TTFT_TIERS.md` "tokenwise stage starves its downstream links" note describes recv-timeout expiry under long compute — a different mechanism (budget, not starvation) — unchanged here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -446,6 +446,7 @@ dependencies = [
  "cascadia-engine",
  "cascadia-engine-mock",
  "cascadia-metrics",
+ "cascadia-transport",
  "cascadia-types",
  "futures",
  "parking_lot",
--- a/crates/cascadia-api/src/lib.rs
+++ b/crates/cascadia-api/src/lib.rs
@@ -1865,7 +1865,10 @@ fn engine_error_response(err: cascadia_engine::EngineError) -> axum::response::R
         EngineError::QueueFull { .. } => "capacity",
         EngineError::PromptTooLong(_) => "prompt_over_window",
         EngineError::NotLoaded | EngineError::NotConnected => "engine_unavailable",
-        EngineError::Backend(_) | EngineError::Io(_) | EngineError::Task { .. } => "engine_error",
+        EngineError::Backend(_)
+        | EngineError::Io(_)
+        | EngineError::Task { .. }
+        | EngineError::BatchAborted(_) => "engine_error",
         EngineError::InvalidConfig(_)
         | EngineError::PeerRejected(_)
         | EngineError::ShardRejected(_)
@@ -1885,6 +1888,9 @@ fn engine_error_response(err: cascadia_engine::EngineError) -> axum::response::R
         // A task-attributed step failure: the engine abandoned a task; the
         // underlying cause is a backend/transport failure.
         EngineError::Task { .. } => StatusCode::INTERNAL_SERVER_ERROR,
+        // A pipeline stage abandoned the in-flight batch. Server-side and
+        // this node's to explain, like any other engine failure.
+        EngineError::BatchAborted(_) => StatusCode::INTERNAL_SERVER_ERROR,
     };
     (status, Json(serde_json::json!({"error": err.to_string()}))).into_response()
 }
@@ -1949,6 +1955,7 @@ mod tests {
             (EngineError::NotLoaded, "engine_unavailable"),
             (EngineError::NotConnected, "engine_unavailable"),
             (EngineError::Backend("boom".into()), "engine_error"),
+            (EngineError::BatchAborted("x".into()), "engine_error"),
             (EngineError::InvalidConfig("bad".into()), "invalid_request"),
             (EngineError::ModelNotFound("nope".into()), "invalid_request"),
         ];
--- a/crates/cascadia-api/src/lib.rs
+++ b/crates/cascadia-api/src/lib.rs
@@ -1132,11 +1132,13 @@ async fn chat_completions(
         trust_remote_code: false,
     };
 
-    // Acquire a request slot before touching the engine. Without this
-    // a flood of concurrent SSE callers would hammer one engine mutex
-    // and starve everyone (the `MAX_CONSECUTIVE_EMPTY_STEPS=3` guard
-    // in the runner would then truncate streams). Backpressure is
-    // 503; clients should retry with backoff.
+    // Acquire a request slot before touching the engine. Contended streams
+    // no longer block a tokio worker (#122: they park on a failed `try_lock`,
+    // and submits go out via `spawn_blocking`), so this gate is not what
+    // keeps the runtime alive — it bounds the two resources that are still
+    // finite: the engine's own queue/slot depth, and the blocking pool the
+    // submits land on. Backpressure is 503; clients should retry with
+    // backoff.
     let permit = match state.permits.clone().try_acquire_owned() {
         Ok(p) => p,
         Err(_) => {
--- a/crates/cascadia-api/src/lib.rs
+++ b/crates/cascadia-api/src/lib.rs
@@ -1183,7 +1183,7 @@ async fn chat_completions(
     // until the task completes; drop frees the slot.
     let _permit = permit;
     let _inflight = inflight;
-    let mut chunk_stream = match state.runner.generate(task.clone()) {
+    let mut chunk_stream = match state.runner.generate_async(task.clone()).await {
         Ok(s) => s,
         Err(err) => return engine_error_response(err),
     };
@@ -1381,7 +1381,7 @@ async fn completions(
 
     let _permit = permit;
     let _inflight = inflight;
-    let mut chunk_stream = match state.runner.generate(task) {
+    let mut chunk_stream = match state.runner.generate_async(task).await {
         Ok(s) => s,
         Err(err) => return engine_error_response(err),
     };
@@ -1476,7 +1476,7 @@ async fn stream_text_completion(
     include_usage: bool,
 ) -> axum::response::Response {
     let task_id = task.task_id.clone();
-    let chunk_stream = match state.runner.generate(task) {
+    let chunk_stream = match state.runner.generate_async(task).await {
         Ok(s) => s,
         Err(err) => {
             warn!(error = %err, "completions-stream: generate failed");
@@ -1599,7 +1599,7 @@ async fn stream_completion(
     let task_id = task.task_id.clone();
     let _ = SystemTime::now();
 
-    let chunk_stream = match state.runner.generate(task) {
+    let chunk_stream = match state.runner.generate_async(task).await {
         Ok(s) => s,
         Err(err) => {
             warn!(error = %err, "ov-stream: generate failed");
--- a/crates/cascadia-cli/src/lib.rs
+++ b/crates/cascadia-cli/src/lib.rs
@@ -1527,23 +1527,13 @@ fn validate_worker_runtime_flags(args: &WorkerArgs) -> Result<()> {
         if args.packed_slots < 2 {
             return Err(anyhow!("--packed-slots must be 0 (off) or >= 2"));
         }
-        // Multi-stage packed is withheld pending a wire fault. The mechanism
-        // works — stage 0 ships an I64 [1,3,S] plan frame (slot, absolute
-        // position and prefix-reuse length per row) ahead of the [1,S,hidden]
-        // block, and the tail replies one token per row — but a token frame
-        // intermittently goes missing between the ranks, and rank 0 then blocks
-        // forever on a reply that never comes. Instrumented on NPU: rank 1
-        // logged the reply and advanced, rank 0 never saw it, no error on
-        // either side. Reproduces after sustained load, roughly two runs in
-        // three. Single-stage is unaffected and verified.
-        if args.total != 1 {
-            return Err(anyhow!(
-                "--packed-slots is single-stage only (--total 1); multi-stage packed can lose a \
-                 token frame between ranks and wedge the pipeline (issue #122). Run the packed \
-                 worker as a single stage, or drop --packed-slots to use the multi-stage baseline \
-                 path"
-            ));
-        }
+        // Multi-stage packed is allowed again: the #122 wedge was driver
+        // starvation on rank 0 (sync engine-mutex blocking pinned every
+        // tokio worker, so the token-frame reply could neither be read nor
+        // timed out), fixed in cascadia-runner (non-blocking poll/submit/
+        // cancel) plus deadlined+poisoning reply recvs and an on-wire NACK
+        // in the packed exchange. Every stage must run the same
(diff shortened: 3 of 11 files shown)


t8:
## NPU rig confirmation — the issue's exact environment

Ran the confirmation pass on the original hardware/topology: **2-box Lunar Lake NPU pipeline** (rank 0 on pawan-02, rank 1 on pawan-03, cross-box over tailscale), this branch's binary (`da8a892c`) built against the OV GenAI **2026.2** Windows SDK (sha256-verified identical on both boxes), and the same 2-stage Llama-3.2-1B packed export from the issue's repro command (`--target npu --static-context 1024 --static-prefill-seq 64 --packed-slots 4`), both workers `--device NPU --packed-slots 4`.

**Sustained-load probe** (the wedge trigger: 15 solo then 6 concurrent, × 4 rounds, `max_tokens=48`):

```
ALL_OK requests=84 wall=269.9s
```

Zero client errors, zero hangs, and **zero WARN/ERROR lines in either rank's log** across the whole run (smoke + probe + capture). The original repro wedged ~2 runs in 3 on the first concurrent batch after sustained solo load; the fixed binary cleared four consecutive such cycles.

**Multi-stage packed accuracy on NPU** — `packed_accuracy.py capture` ran to completion (this is the capture #122 reported as un-obtainable because it wedged mid-run), 10 prompts × 3 passes × 128 tokens:

| check | result |
|---|---|
| determinism (solo vs solo-repeat) | **10/10 exact** |
| batch-composition invariance (solo vs batched) | **10/10 exact** |
| teeth | 10/10 distinct outputs, completion_tokens up to 128 |

Better than both the CPU-device run in the PR body (9/10 invariance) and the issue's single-stage control (8/10).

Fleet state: workers stopped and the temporary firewall rule removed; the built exe (`C:\cascadia\cascadia122.exe`), OV 2026.2 runtime DLLs (`C:\cascadia\ov262`), and the packed export (`C:\cascadia\models\llama1b-2stage-packed`, stage-split across the two boxes) are left in place for future NPU validation. Capture JSON: `mstage2_packed_npu.json` on pawan-02.

7i7o (approved):
tACK 👍

PR #132, opened 2026-08-13
feat(cli): --elastic memory posture + --ov-config passthrough

t8:
## Summary

Two memory-oriented `worker` flags for the OpenVINO path.

**`--ov-config KEY=VALUE`** (repeatable) — raw OpenVINO plugin-property passthrough, merged into the property set alongside the typed `--ov-*` flags and applied **last** (override-wins). Parse-validated (`KEY=VALUE`, non-empty key); NPU-prefixed keys reuse the existing NPU device + `ov-genai` gate. The escape hatch for plugin knobs without a dedicated flag — e.g. `--ov-config ENABLE_MMAP=YES`, `--ov-config KV_CACHE_PRECISION=u8`. Cross-platform.

**`--elastic`** (+ `--elastic-min-mb` / `--elastic-pool-mb`) — serve large allocations from file-backed mappings so the worker's weight copies, KV and scratch become kernel-reclaimable instead of private/committed. New `cascadia-elastic` crate:

- **Linux** — embeds a compiled interposer, writes it to a private temp file, and `LD_PRELOAD`s it via a one-shot guarded re-exec (serving PID unchanged via `execv`).
- **Windows** — inline-hooks the UCRT allocation family (`malloc`/`free`/`realloc`/`calloc`/`_msize` + the `_aligned_*` variants) with Microsoft Detours, in-process, before the OV engine loads. Compiled + linked only when `DETOURS_DIR` is set at build time (an SDK-style input, like `INTEL_OPENVINO_DIR`); without it the hook is compiled out and `--elastic` reports inactive.

Activation runs in `main` before the async runtime starts (env mutation + `execv` are only safe single-threaded).

## Measured — unmodified OpenVINO CPU worker

| platform | engine / model | stock | `--elastic` | delta |
|---|---|---|---|---|
| Linux (Cascade Lake) | ov-runtime, 8B-class int4 | 2064 MB anon | 506 MB | 4.1×, −1% decode |
| Windows (Lunar Lake) | ov-genai, Qwen3-1.7B-int4 | 1329 MB private | 223 MB | 5.95×, −6% decode |

Output byte-identical; working set unchanged (the weight pages stay resident, now file-backed). `--elastic-min-mb 16` (weights-only threshold) gives 2064→1549 MB at speed parity. Standalone Windows microbench: 300 MB `malloc` + 300 MB `_aligned_malloc` kept out of private commit (+1 MB vs +301).

## Scope / caveats (for review)

- The interposer reduces **committed / private** memory (the multi-tenancy + OOM-exposure metric), **not** physical residency — it file-backs the engine's weight copy, it does not prevent the copy. OpenVINO exposes no property to disable oneDNN's repack, and a warm model-cache import does not mmap it clean either. Collapsing to a single clean copy would need a targeted plugin change.
- CPU-plugin serving; GPU/NPU device memory (USM) bypasses the CRT heap.
- Windows requires a pre-built Detours via `DETOURS_DIR`. **Vendoring Detours vs. keeping the env-var build input is an open decision — feedback welcome.**

## Build / test

- Builds with `--features openvino` on Linux (miner) and Windows/MSVC (Lunar Lake).
- Tests: `cascadia-elastic` (3), `cascadia-cli --ov-config` (6). Linux release build verified unchanged after the Windows additions.
- `docs/CLI.md` updated for both flags.

---

Opening as **draft** for early feedback on the approach — in particular the Windows Detours build input (vendor vs `DETOURS_DIR`) and the commit-vs-physical scope.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ members = [
     "crates/cascadia-api",
     "crates/cascadia-dashboard",
     "crates/cascadia-cli",
+    "crates/cascadia-elastic",
     "crates/cascadia",
     "tests-e2e",
 ]
@@ -46,6 +47,7 @@ cascadia-download = { path = "crates/cascadia-download" }
 cascadia-api = { path = "crates/cascadia-api" }
 cascadia-dashboard = { path = "crates/cascadia-dashboard" }
 cascadia-cli = { path = "crates/cascadia-cli" }
+cascadia-elastic = { path = "crates/cascadia-elastic" }
 
 # Async runtime
 tokio = { version = "1", features = ["full"] }
--- a/crates/cascadia-cli/Cargo.toml
+++ b/crates/cascadia-cli/Cargo.toml
@@ -33,6 +33,7 @@ cascadia-api = { workspace = true }
 cascadia-dashboard = { workspace = true }
 cascadia-discovery = { workspace = true }
 cascadia-topology = { workspace = true }
+cascadia-elastic = { workspace = true }
 tokio = { workspace = true }
 axum = { workspace = true }
 hyper = { version = "1", features = ["server", "http1"] }
--- a/crates/cascadia-cli/src/lib.rs
+++ b/crates/cascadia-cli/src/lib.rs
@@ -357,11 +357,11 @@ pub struct WorkerArgs {
     /// mappings so the engine's weight copies, KV state and scratch become
     /// kernel-reclaimable instead of anonymous/committed. Measured in-tree
     /// (ramlab exp 199): 2064→506 MB committed at −1% decode on an unmodified
-    /// OpenVINO CPU worker. LINUX ONLY today — it re-execs the worker once with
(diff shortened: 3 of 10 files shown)


SergiioB:
Pushed to your branch per your go-ahead (head bedfd566, two commits) plus #166 with the test suite:

- the interposer now also catches large anonymous mmap calls — oneDNN scratch bypassed the malloc hook (measured ~320 MB staying anon on gemma-26b); those mappings are served from the same O_TMPFILE pool, fd closed after mmap, and the retention pool no longer holds fds (fixes EMFILE at low thresholds)
- measured on B70: gemma-26b settled floor 527 → 172 MB, output byte-identical, and the two-model co-tenancy leg passes under a 500 MB fleet cap (was 1600)
- known trade-offs, both narrow: the mmap leg swaps anon-private for MAP_SHARED (no COW after fork — fine for serving, flagged for review), and MADV_DONTNEED on the file-backed mappings no longer drops dirty pages
- #166 README has the full numbers

PR #134, opened 2026-08-14
fix(dashboard): honest / + embed the SPA in release bundles

t8:
## Problem

A user built main per the README (`cargo build --release -p cascadia`), started a worker with `--api`, saw `API + dashboard serving` in the log — and got an empty-body `404` on `/`:

```
< HTTP/1.1 404 Not Found
< content-length: 0
```

Root cause: the dashboard SPA is only compiled in behind the default-off `dashboard-embed` feature, and **no documented build path enables it** — not the README/QUICKSTART/INSTALL commands, and not the release workflow either, so even release bundles shipped without the UI. The startup log claimed a dashboard unconditionally, and the docs never mentioned one (the log line was the only hint it existed). Bonus trap: `web/dist/` is gitignored, so even `--features dashboard-embed` on a fresh checkout died in three cryptic `E0599` errors from `#[derive(Embed)]` unless you knew to run npm first.

## Fix (three parts)

1. **Honest serving** (`fix(dashboard)`): without the embed, `/` now serves a small built-in pointer page — live endpoints + the exact npm/cargo steps — instead of a bare 404; other unknown paths still 404 (no client routes to serve). The startup log only says "dashboard" when the build embeds one, keyed off the new `cascadia_dashboard::SPA_EMBEDDED` const (evaluated in the feature-owning crate, so feature unification can't make the log disagree with the router). A new build script turns the missing-`web/dist` compile errors into an actionable message.
2. **Release bundles get the dashboard** (`ci(release)`): both release jobs run `npm ci && npm run build` in `crates/cascadia-dashboard/web` (runner-preinstalled Node; JS deps pinned via `package-lock.json`) and build with `--features openvino,dashboard-embed`. If the npm step is ever dropped, the build script fails the job instead of silently shipping a dashboard-less bundle.
3. **Docs** (`docs`): README gains a "Web dashboard" section (what it is, why plain cargo builds serve a pointer page, the two-step embed build, the `vite dev` flow), QUICKSTART points at it from the first curl, INSTALL covers the feature for both build modes, ARCHITECTURE.md one-liner updated.

## Testing

Unit: two new router tests (pointer page served at `/` with the feature off; unknown paths still 404). `cargo test -p cascadia-dashboard` (4 passed), `cargo test -p cascadia-cli` (79 passed), `cargo fmt --check`, `cargo clippy --workspace --all-targets` clean for the touched crates.

E2E, macOS (debug) and the miner (Linux, release — fresh shallow clone of the repo, same environment class as the report):

- **Default build** (`cargo build --release -p cascadia`): log now reads `API serving (dashboard UI not embedded in this build; GET / explains how to enable it)`; `GET /` → 200 pointer page; `/nope` → 404; `/health`, `/api/stats`, `/api/topology` unchanged.
- **`--features dashboard-embed` without `web/dist`**: build stops with the instructions message (was: 3 × E0599).
- **`--features dashboard-embed` after `npm ci && npm run build`**: log reads `API + dashboard serving`; `GET /` serves the SPA; client route `/chat` → SPA shell; hashed `/assets/*.js` served with correct MIME; reserved `/v1/*` misses still 404. On the miner, `web/dist/` was then **deleted and the binary re-run**: the SPA still serves, proving the release binary truly embeds the assets (single-static-binary claim).

The workflow change itself can only fully run on the next `v*` tag; YAML validated, and the npm build step was exercised locally and on CI-equivalent inputs (`npm ci` from the committed lockfile).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,3 +69,24 @@ jobs:
         # clean runtime error. Runs every unit test plus the binary
         # e2e tests against the mock engine.
         run: cargo test --workspace --all-targets
+
+      # Everything above builds default features, and `dashboard-embed` is
+      # default-off — so without these steps the whole embedded-SPA path
+      # (spa.rs, rust-embed, mime_guess, and build.rs's success branch) is
+      # first compiled by the release workflow, at tag time. That workflow
+      # only runs once release-please has already created the tag and the
+      # GitHub pre-release, so a break there strands a published release
+      # with no binaries attached. Compile it on every PR instead.
+      #
+      # Runner-preinstalled Node (>= 20 on all hosted images); the JS deps
+      # are pinned by package-lock.json via `npm ci`.
+      - name: Build dashboard SPA
+        working-directory: crates/cascadia-dashboard/web
+        run: npm ci && npm run build
+
+      - name: Lint + test (dashboard embed path)
+        # Also the only lint spa.rs ever gets: the clippy run above is
+        # default-features, so the file is invisible to it.
+        run: |
+          cargo clippy -p cascadia-dashboard --features embed-spa --all-targets
+          cargo test -p cascadia-dashboard --features embed-spa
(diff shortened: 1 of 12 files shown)


7i7o (approved):
tACK 👍

PR #137, opened 2026-08-20
feat(engines): forced-prefix resume + sparse-moe per-token streaming

pawanpaudel93:
## Summary

Engine-side support for **forced-prefix resume**: a scheduler that re-dispatches a mid-stream failure to an alternate chain sends the already-emitted token ids as `resume_token_ids`; engines append them to the rendered prompt, seed the `generated` accumulator, force greedy decode, and never re-emit the prefix. Also converts the two remaining burst engines to per-token streaming so a mid-stream window exists on every family, and makes the native K2.6 engine a fully servable chain participant (head IR export + cross-chain KV carry).

### Types (`cascadia-types`)
- `GenerationTask.resume_token_ids: Option<Vec<i32>>` — serde-additive; old payloads decode unchanged.
- `resume_ids()` normalizer (wire-legal `Some([])` = not resuming), `validate_resume_ids` (sign + vocab bounds — peer-supplied ids reach native gathers), `append_resume_ids` / `resume_generated_seed` helpers.

### Resume support in every engine family
- **ov-runtime, gemma4, qwen36, dist-spec, sparse-moe (K2.6), MiniMax-M2**: admission validation; budget pre-check (seed ≥ budget → zero-token `Length` final before any decode); KV-capture keys skip the seed; seam handling per path (whole-text re-anchor on burst, seeded byte-cursor on streamed, U+FFFD hold on gemma4, shim `advance_emitted` on qwen36).
- **Mock engine** enforces id-fidelity (seed id at position *i* must be *i*) with a negative test — pins that engines consume the ids, not just their count.

### Per-token streaming for the burst engines
- `SparseMoEEngine` + `OvMoeEngine` multi-stage rank-0 converted from whole-turn burst to `begin/decode/finalize` state machines (same scope `PipelineEngine` shipped; single-stage and spec-decode stay burst, and a resumed turn routes to the streamed driver).
- Mid-decode failure emits an **error chunk**, never a partial-success final — a final marker would latch the stream as complete and block any downstream rescue.
- `token_ids = vec![id]` stamped on every streamed chunk (including `PipelineEngine`) so a legal token-0 sample can't read as id-less.
- `cancel()` clears in-flight state; empty-prompt/`-1` guards; history `mem::take` (was an O(n²) per-token clone).

### Native K2.6 as a servable chain (head IR + cross-chain KV)
- `export_kimi_k26.py --head` writes the RMSNorm + lm_head OV IR a last stage needs — the historical "native can't be a last stage" limitation was only the missing head file. Input re-typed to **bf16** via PrePostProcessor (the runner sets input 0 as bf16; an f32-traced IR fails OV `set_tensor` with `ParameterMismatch` and wedges the head on its single admission slot).
- **Capture epoch keyed over the KV-covered prefix** (`full_tokens[..past_seq_len]`): the native engine's test-before-push EOS semantics leave the final sampled token un-pushed, so the turn's token vector runs one longer than the KV — an epoch over the full vector never matches a later NEGOTIATE's offer epoch and every worker GET answers `get_none`. The OvMoe engine's push-then-test semantics made the two vectors coincide, which is why this only surfaced on the native path. Mirrors the OvMoe capture site, which already truncated.
- **Native carried-slice restore (`RESTORE_CARRY`)**: `stash_downstream_rank` was the trait default `Err(())` — every rank>0 cross-chain pull voted cold. Added the native `KvSnapshot` blob codec, the carry stash (with the same epoch-collision guard as OvMoe), a carry-aware `forward_restore_downstream`, and the worker-side handler (event `sparse_tail_restore_carried`).
- **Donor-side serve diagnostics**: the holder's refusal arms were silent — a puller's `pull_not_found` was indistinguishable from a missing capture, a length drift, or a namespace mismatch. Added `kv_capture_stored`, `kv_serve_capture_miss` (with held epochs), `kv_serve_len_drift`, and a warning when a capacity-0 worker acks a capture it silently skipped.

### Declines
- Paths that cannot honor a forced prefix (MiniMax single-stage) decline with a `resume_unsupported:`-prefixed error as their first chunk, so the caller can exclude and retry.

## How to test

```bash
cargo test --workspace                                       # full suite
cargo test -p cascadia-engine-sparse-moe --features kv_coord # feature lane
cargo test -p cascadia-engine-sparse-moe --test prepare_resume
```

Notes:
- Engine-level streaming/resume integration tests (`tests/sparse_streaming.rs`, `tests/ovmoe_streaming.rs`) are env-gated on model directories (`K26_TINY_DIR` / `M2_MODEL_DIR`) and skip with a message when unset; the hardware cert is the enforcement gate for those paths.
- Per-engine resume unit tests (seed consumption, budget, capture-key skip, seam) live beside each engine and run in the default suite; the mock engine's id-fidelity check has a negative test (`resume_with_wrong_ids_is_rejected`).

## Verification

- Full workspace suite green; sparse-moe crate green on default and `kv_coord` feature lanes.
- Hardware rig (four nodes, two 2-stage chains): resume-cert **PASS on all six engine cells** — ov-runtime 5/5, dist-spec 5/5, qwen36 5/5, gemma4 5/5, MiniMax-M2 sparse-moe 4/4, native K2.6 sparse-moe 4/4 — serving tail killed mid-stream, rescue via forced prefix, no replay fallback, stream reaches `[DONE]`.
- Native K2.6 cross-chain KV move-cert **7/7 bars**, genuinely cross-chain: pull-hit counter +1, tail restored from the head's carried blob (+1), `plane_pulled=true`, and the pulled-KV continuation byte-identical to a cold re-prefill.

### Final-review fix wave

A two-reviewer full-branch pass added: sentinel declines on the remaining resume-incapable paths (`PipelineEngine::begin_generation`, both ov-genai submits — silently ignoring a forced prefix regenerates from scratch and the caller splices it as a fake resume), a one-shot seam-divergence guard on both streamed resume paths (mirror of the burst path's), a dims cross-check + 4 codec tests for the carry blob, verdict-mirror and warn-parity fixes on the carry/capture paths, and comment hygiene. Re-certified on hardware after the wave: resume cert 5/5, plus the native-K2.6 cells whose binaries gained active-path code — KV move-cert 7/7 (first attempt) and resume cert 4/4.

## Deliberate non-goal

**`StagedRunner` unification** (one streaming engine for K2.6 / M2 / glm5 / dsv4) is analyzed and rejected for this PR: the trait's `&self` receivers conflict with OpenVINO's `&mut infer()` (fixing it breaks 12 external call sites), `max_seq()` has no source of truth, and the two wire dialects mutually reject each other's frames. ~4.4k lines of regression risk across certified engines for zero functional gain — every family already streams and resumes. Full analysis in the downstream design spec.

### DIFF
--- a/crates/cascadia-api/src/lib.rs
+++ b/crates/cascadia-api/src/lib.rs
@@ -1839,6 +1839,10 @@ async fn chat_completions(
         // client that could set it from request JSON could name another tenant's namespace and
         // resume off its prefix. The real value is assigned server-side from the admitted identity.
         tenant: String::new(),
+        // Never client-settable, same class as `tenant`: a client that could
+        // seed it from request JSON would inject arbitrary token-ids into the
+        // model's context PAST the rendered prompt. Only the scheduler's
+        // rescue path sets it, engine-side of this API.
         resume_token_ids: None,
     };
 
@@ -2056,6 +2060,7 @@ async fn completions(
         trust_remote_code: false,
         // H.1b: hardcoded server-side; see the note on the chat endpoint's task above.
         tenant: String::new(),
+        // Never client-settable; see the chat endpoint's note.
         resume_token_ids: None,
     };
 
--- a/crates/cascadia-api/src/lib.rs
+++ b/crates/cascadia-api/src/lib.rs
@@ -1839,6 +1839,7 @@ async fn chat_completions(
         // client that could set it from request JSON could name another tenant's namespace and
         // resume off its prefix. The real value is assigned server-side from the admitted identity.
         tenant: String::new(),
+        resume_token_ids: None,
     };
 
     // Acquire a request slot before touching the engine. Contended streams
@@ -2055,6 +2056,7 @@ async fn completions(
         trust_remote_code: false,
         // H.1b: hardcoded server-side; see the note on the chat endpoint's task above.
         tenant: String::new(),
+        resume_token_ids: None,
     };
 
     let permit = match state.permits.clone().try_acquire_owned() {
--- a/crates/cascadia-cli/src/lib.rs
+++ b/crates/cascadia-cli/src/lib.rs
@@ -2033,6 +2033,7 @@ async fn cmd_worker(args: WorkerArgs) -> Result<()> {
             enable_thinking: false,
             trust_remote_code: false,
             tenant: String::new(), // H.1b: single-tenant CLI path — LOCAL_NS
+            resume_token_ids: None,
         };
         // `generate_async`, not `generate`: this loop runs on the tokio
         // runtime, and the sync path would block a worker on the engine
--- a/crates/cascadia-engine-mock/src/lib.rs
+++ b/crates/cascadia-engine-mock/src/lib.rs
@@ -92,7 +92,8 @@ impl Engine for MockEngine {
             )]);
         }
         let token = words[emitted].to_string();
-        let chunk = Chunk::token(&task.task_id, emitted as i64, token + " ");
+        let chunk =
+            Chunk::token(&task.task_id, emitted as i64, token + " ").with_token_ids(vec![emitted as i64]);
         let task_id = task.task_id.clone();
         self.pending.push((task, emitted + 1));
         Ok(vec![(task_id, chunk)])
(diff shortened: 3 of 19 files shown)


7i7o (approved):
tACK 👍

PR #138, opened 2026-08-20
ci(ai-pc): fix Windows self-hosted reliability

twilson63:
Ensures Git bash is on PATH before dtolnay/rust-toolchain on the self-hosted Windows ai-pc runner.

Self-hosted Windows has bash at `C:\Program Files\Git\bin\bash.exe` but the runner service PATH does not include it, causing `bash: command not found` at the rust-toolchain parse step. Adds an Ensure bash step that appends Git bin to GITHUB_PATH and fails clearly if Git for Windows is not installed.

Fixes the ai-pc runner setup for ai-pc-tests.yml.


7i7o (approved):
tACK 👍

PR #140, opened 2026-08-21
fix(dsv4): read CASCADIA_DSV4_PREFIX_CACHE for the prefix cap, not the glm5 knob

### LINKED ISSUE #139: dsv4: prefix cache reads the glm5-named env knob (CASCADIA_GLM5_PREFIX_CACHE) → cross-arch tuning bleed
Pre-existing; surfaced during the kimi-k3 review when the k3 arm got its own `CASCADIA_K3_PREFIX_INDEX`. Deliberately preserved there (the k3 change was scoped to k3), so filing it as its own fix.

## What happens

`crates/cascadia-engine-sparse-moe/src/engine.rs`, `Builder::build`, dsv4 arm (~line 830):

```rust
// dsv4 has no config-threaded prefix cache; `None` keeps its
// existing env-only behaviour byte-for-byte, including reading
// the glm5-named var (out of scope here — dsv4 never had its
// own knob, so this preserves rather than fixes that gap).
None,
"CASCADIA_GLM5_PREFIX_CACHE",
```

The dsv4 engine's prefix-cache knob is the *glm5*-named env var. The in-code comment acknowledges the gap.

## Two consequences

**1. Undiscoverable knob.** An operator tuning dsv4's prefix cache has to set a GLM5-named variable that no dsv4 doc, error message, or naming convention would lead them to.

**2. Cross-arch bleed.** Setting `CASCADIA_GLM5_PREFIX_CACHE` for an actual glm5 node silently retunes any dsv4 engine sharing the environment. `resolve_prefix_cap`'s own doc-comment (~line 3010) says the name-parameterization exists so "an arch-scoped knob ... can never reach a different arch's engine" — the dsv4 arm is the one caller that defeats it.

## Naming standard (settled — not up for relitigation here)

`CASCADIA_<ARCH>_<KNOB>` is the convention for engine env knobs; this arm is the only deviation. The env layer stays arch-scoped — knob units differ per arch (glm5's prefix knob is a depth, k3's cache is bytes, k3's index cap is entries), and a unified name reintroduces the bleed above. Operator-facing unification belongs to the `engine_args` config layer (cf. the `max_seq` alias), not env names.

## Fix

- Introduce `CASCADIA_DSV4_PREFIX_CACHE`; the dsv4 arm reads it first.
- Fall back to `CASCADIA_GLM5_PREFIX_CACHE` when unset, with a deprecation `warn!` naming both vars, so existing deployments keep working through one release.
- Drop the fallback after the deprecation window.

`resolve_prefix_cap` is already name-parameterized, so the change is confined to the dsv4 arm's argument plus the fallback shim.

## Verification

Unit-testable without hardware: mirror the existing `resolve_prefix_cap` / `prefix_index_cap` regression tests — new var wins, old var honored-with-warn when new is unset, neither set keeps today's default.

pawanpaudel93:
## Summary

Closes #139

The dsv4 engine's rank-0 KV-prefix cap was resolved inside the shared `PipelineEngine::new` by reading the glm5-named `CASCADIA_GLM5_PREFIX_CACHE` whenever the caller passed `None` — and the dsv4 arm always passed `None`. That let a glm5 tuning reach a dsv4 engine sharing the environment, and made the dsv4-scoped name nonexistent.

**Changes**

- `engine.rs`
  - `resolve_prefix_cap(explicit, env_var)` — explicit wins, else reads the *named* var; unset/unparseable → `0` (off). The glm5 arm passes `"CASCADIA_GLM5_PREFIX_CACHE"`; its behaviour is byte-identical to before.
  - `dsv4_prefix_cap(dsv4_raw, glm5_raw)` — `CASCADIA_DSV4_PREFIX_CACHE` wins whenever set (set-but-empty counts as set → off). While it is unset, the legacy glm5 name is honoured with a deprecation `warn!` naming both vars. The shim is one function; drop it after the deprecation window.
  - `PipelineEngine::new` now takes the already-resolved `usize`, so no env read lives in the shared constructor.
- Unit tests (pure / test-unique env name, no cross-test races): new var wins, legacy honoured when new unset, neither → `0`, unparseable new var does not fall back, empty new var is off not fallback, whitespace trimmed; `resolve_prefix_cap` explicit/env/trim/unparseable/unset.

**Scope note — the cap is currently inert for dsv4**

`Dsv4Runner` implements none of the `StagedRunner` prefix hooks (`prefix_cache_enabled` / `restore_prefix` / `cache_prefix` / `supports_batched_prefill` all default to off, `staged.rs`), and every use of `prefix_cap` in the engine is gated on them. So on `main` the bleed in #139 is latent, not live; this PR is the naming/plumbing fix so that it cannot become live when dsv4 grows prefix hooks (or under #121's store-aware default). For that reason the knob is deliberately **not** documented in `docs/architectures/deepseek-v4.md` yet — the PR that wires dsv4 prefix hooks should document it. The dsv4 arm comment states this.

**Verification**

All pass locally (`just check` by-hand equivalent; the e2e crate needs the binary built first):

```sh
cargo fmt --all -- --check
cargo clippy --workspace --all-targets
cargo build -p cascadia
cargo test --workspace --all-targets
```

Focused run for this change (`cascadia-engine-sparse-moe` lib: 101 passed, 7 new; `glm5_prefill_pipeline`, which sets the glm5 var, still green):

```sh
cargo test -p cascadia-engine-sparse-moe --lib prefix_cap
```

CI: stub + cargo-deny green. The ai-pc OpenVINO lane fails with `STATUS_DLL_NOT_FOUND` on the Windows self-hosted runner — same failure on `main` @ `1ea79b8c`, unrelated to this diff.

**Merge shape vs #121 (kimi-k3)**

Both PRs define `resolve_prefix_cap` and change `PipelineEngine::new`. This PR keeps the *caller-resolves* shape (`new` takes a `usize`; no env read in the shared ctor — the root cause named in #139). #121 resolves inside `new` with a 3-arg `resolve_prefix_cap(explicit, env_var, store_enabled)`. Whichever lands second: keep caller-resolves, hoist #121's `resolve_prefix_cap(.., runner.prefix_cache_enabled())` into the build arms (where `runner` is already a local), and pass the resolved `usize`. Conflict is the helper body, the two arm arguments, the `new` signature, and the tests tail — mechanical.

**Follow-up**

- Remove the glm5 fallback + `warn!` after one release.

## Checklist

- [x] `just check` passes locally (see [CONTRIBUTING.md](https://github.com/labscommunity/cascadia/blob/main/CONTRIBUTING.md) for the by-hand equivalent)
- [x] Commits follow [Conventional Commits](https://www.conventionalcommits.org/) (`feat:`, `fix:`, `docs:`, …), one logical change per commit
- [ ] If this touches an OV engine: tested on real Intel hardware (say which, and which OpenVINO version) — CI only covers stub mode — N/A: sparse-moe env resolution only, no OV engine code touched; not hardware-tested

### DIFF
--- a/crates/cascadia-engine-sparse-moe/src/engine.rs
+++ b/crates/cascadia-engine-sparse-moe/src/engine.rs
@@ -720,9 +720,11 @@ impl Builder for SparseMoEBuilder {
                 runtime_handle,
                 rank,
                 total,
-                // dsv4 has no config-threaded prefix cache; `None` keeps its
-                // existing env-only behaviour byte-for-byte.
-                None,
+                // dsv4 has no config-threaded prefix cache; env-only.
+                dsv4_prefix_cap(
+                    std::env::var("CASCADIA_DSV4_PREFIX_CACHE").ok().as_deref(),
+                    std::env::var("CASCADIA_GLM5_PREFIX_CACHE").ok().as_deref(),
+                ),
             )));
         }
         if let Some(runner) = self.glm_runner {
@@ -745,7 +747,10 @@ impl Builder for SparseMoEBuilder {
                 total,
                 // Same value the per-rank SliceKvCache got, so the rank-0 index
                 // and the per-rank caches stay in lockstep.
-                self.config.prefix_cache_depth.map(|d| d as usize),
+                resolve_prefix_cap(
+                    self.config.prefix_cache_depth.map(|d| d as usize),
+                    "CASCADIA_GLM5_PREFIX_CACHE",
+                ),
             )));
         }
         if let Some(ov) = self.ov_runner {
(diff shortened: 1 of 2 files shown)


pawanpaudel93:
Closing — superseded by labscommunity/cascadia@ccec9908 on `pawan/kimi-k3` (#121).

On closer review dsv4 has no prefix cache: `Dsv4Runner` implements none of the `StagedRunner` prefix hooks (`supports_batched_prefill` / `prefix_cache_enabled` / `restore_prefix` / `cache_prefix`; trait defaults in `staged.rs` are off, and Rust has exactly one impl per type), and every reader of `prefix_cap` in `PipelineEngine` is gated on them. For dsv4 the cap is written once and never read — a dead store — so `CASCADIA_DSV4_PREFIX_CACHE` would be a knob that does nothing, and there is no existing behaviour for a fallback to preserve.

The one real item — the shared `PipelineEngine::new` reading an arch-named env var for whatever runner it is handed — is now fixed on #121, which owns that constructor: each build arm resolves its own cap with its own var and `new` takes the resolved `usize`; the dsv4 arm passes `0`.

Branch kept locally for reference.

Issue #144, opened 2026-08-22
Elastic runtime rebalance of expert-cache / pin / KV budgets at safe points


t8:
## Motivation

Every memory budget in Cascadia is fixed at launch:

- K2.6 expert-LRU capacity: resolved once from env/config at load (`crates/cascadia-engine-sparse-moe/src/engine.rs`).
- GLM/DSV4 pin budget: computed once from `mem_available()` at load (`glm/residency.rs`).
- ov-runtime KV window: baked into the exported IR (`--static-context`); ov-genai CB cache: fixed `--cb-cache-size`, exhaustion is a task failure.

The only live reaction to memory pressure on main is the Windows working-set trim governor (`glm/stage.rs`), which can shrink but never re-grow, and touches nothing but the working set. Meanwhile the machines are shared — compositor, browser, other stages on small-RAM fleet boxes — so the memory actually available mid-run routinely diverges from what load-time saw, in both directions: pressure causes thrashing against a too-large pin set, and freed memory goes unused by a too-small one.

FreeToken (*Efficient Edge-Native MoE Serving with Bandwidth-Adaptive Execution*, Yang et al., [arXiv:2608.16157](https://arxiv.org/abs/2608.16157)) — the paper this opportunity comes from — treats this as a first-class mechanism: at scheduler **safe points** it rebuilds the expert cache under a revised memory budget **without restarting the engine or reloading the host-resident pool**, re-splitting free memory between the KV cache (grows with context) and the expert working set (stable) per turn.

## Proposal

- Re-sample `mem_available()` periodically (the sampling helper and the 20 s governor cadence already exist).
- At safe points (between decode steps / between tasks), re-run the `pin_expert_count` arithmetic and grow/shrink the GLM pin set and the K2.6 LRU cap to the revised budget — shrink under pressure before the OS starts evicting for us, re-grow when memory frees.
- Longer term: fold KV growth into the same budget (KV demand rises over a long generation; the expert budget should yield to it, which is FreeToken's KV-vs-expert split on our UMA terms).

## Relation

This is the runtime-budget half of the elasticity story; the allocator-posture half (`--elastic` host-memory posture, `--ov-config` passthrough) is on the `feat/elastic-ov-config` branch. The two compose: the allocator makes memory returnable, this issue makes the budgets follow it.

Credit: opportunity identified while reviewing **FreeToken, arXiv:2608.16157** against the Cascadia codebase.

Issue #143, opened 2026-08-22
Recency-adaptive expert pin rotation for the GLM/DSV4 mmap path


t8:
## Motivation

The GLM-5.2 / DeepSeek-V4 expert pin set is **frequency-static**: at load, `pin_expert_count` sizes a budget and the hottest experts from the persistent `.coli_usage` selection histogram are `mlock`ed (`crates/cascadia-engine-sparse-moe/src/glm/residency.rs`, `glm/stage.rs`). The set never changes during a run; adaptation only happens across restarts as the histogram accumulates.

FreeToken (*Efficient Edge-Native MoE Serving with Bandwidth-Adaptive Execution*, Yang et al., [arXiv:2608.16157](https://arxiv.org/abs/2608.16157)) — the paper this opportunity comes from — shows that **recency**-based expert caching tracks the model's evolving routing much better than static/frequency policies: at equal capacity their global LRU misses 16% of decode expert reads vs 41% (KTransformers) and 62% (llama.cpp), and hit-rate is the dominant term for streaming-bound decode. Routing distributions drift with the workload (coding vs chat vs long-context), which a boot-time frequency snapshot can't follow. Notably, K2.6's expert cache in this repo is already a global recency LRU (`src/runner.rs`); the GLM/DSV4 mmap path is the outlier.

## Proposal

- Maintain a decayed recent-selection counter per `(layer, expert)` (the `UsageStats` recorder already sees every selection).
- At periodic safe points (every N tokens, between forwards), rotate the pin set within the existing budget: `munlock` the coldest pinned experts, `mlock` the hottest unpinned ones, hysteresis-bounded so churn stays far below NVMe re-read cost.
- Keep the existing interactions intact: the Windows working-set governor (`CASCADIA_GLM5_TRIM_MB`), the `reserve_lockable` min-working-set dance, and the pin/OV-offload mutual exclusion on UMA.
- Validate with the existing `CASCADIA_GLM5_PROFILE` residency hit% metric: the claim to beat is static-pin hit% on a workload whose routing drifts mid-run.

Credit: opportunity identified while reviewing **FreeToken, arXiv:2608.16157** against the Cascadia codebase.

Issue #142, opened 2026-08-22
Semantic-anchored prefix/KV checkpoints so agentic context edits don't force full re-prefill


t8:
## Motivation

Cascadia's prefix/KV reuse is token-exact and structurally blind to conversation shape. On main there are three disjoint mechanisms:

- **NPU packed shared prefix** (`--packed-prefix`, `crates/cascadia-engine-openvino/src/packed_exec.rs`): longest-common-prefix over raw token ids, a **single** cache entry populated by the first admitted request, permanent for process lifetime, no eviction, single-stage only.
- **`--cb-prefix-caching`**: a passthrough to OV GenAI's paged-attention block cache (`crates/cascadia-ov-genai-shim`), granularity owned upstream.
- **sparse-moe `--kv-prefix-cache-size`** (`crates/cascadia-engine-sparse-moe/src/kv_prefix_cache.rs`): LRU of whole post-prefill KV snapshots keyed by exact token-prefix hash.

Nothing anywhere inspects chat-turn, tool-call, or thinking-block boundaries — and nothing can, because chat templates are rendered before the engine sees a flat id stream. Agentic clients constantly **edit** context (drop thinking blocks, truncate tool outputs), which breaks exact-prefix matching at the first edited byte and forces a full re-prefill of everything after it.

FreeToken (*Efficient Edge-Native MoE Serving with Bandwidth-Adaptive Execution*, Yang et al., [arXiv:2608.16157](https://arxiv.org/abs/2608.16157)) — the paper this opportunity comes from — anchors recurrent-state checkpoints at **semantic boundaries** (thinking segments, tool calls, tool outputs, conversation turns). A new request restores from the deepest checkpoint whose position **survives the edit**, so only the genuinely new suffix is re-prefilled. Their agentic-workload result: 12% throughput variance single-turn → multi-turn, vs a 31% first-multi-turn loss for the best baseline.

## Proposal

1. **Plumb boundary metadata**: the API layer knows message/turn/tool-call offsets pre-templating; carry anchor token-offsets alongside the prompt ids down to engines (wire + engine trait plumbing).
2. **sparse-moe**: checkpoint KV snapshots at anchors instead of only once post-prefill; on a new request, restore from the deepest surviving anchor (`kv_prefix_cache.rs` already has the LRU + snapshot machinery).
3. **NPU packed prefix**: allow multiple LRU-managed shared-prefix entries anchored at turn boundaries, replacing the current one-shot permanent entry.
[...]

Issue #141, opened 2026-08-22
Stream layer L+1's expert set during batched prefill — overlap NVMe reads with compute


t8:
## Motivation

During batched prefill, the sparse-MoE engines already amortize expert reads within a layer — `MoeLayer::forward_batch` batch-unions a 128-row block so each unique routed expert's int4 bin is read once (`crates/cascadia-engine-sparse-moe/src/glm/moe.rs`). But there is **no overlap between layers**: layer `l+1`'s expert bins are only read when the block reaches it, so on the NVMe-streaming models (GLM-5.2, DeepSeek-V4, K2.6) prefill alternates between compute-bound and read-bound phases and the disk sits idle while a layer computes.

FreeToken (*Efficient Edge-Native MoE Serving with Bandwidth-Adaptive Execution*, Yang et al., [arXiv:2608.16157](https://arxiv.org/abs/2608.16157)) — the paper this opportunity comes from — double-buffers full layers during prefill: while layer `l`'s routed experts compute, a dedicated transfer stream loads layer `l+1`'s complete expert set. They measure a **26% prefill-throughput penalty at 16k tokens when the overlap is disabled**.

## Proposal

Extend the existing `glm5-lookahead` worker (`src/glm/lookahead.rs`), which already does exactly this shape of work for decode (async sequential warm of predicted next-layer experts, residency-gated, stale-job dropping):

- When `forward_batch` starts layer `l`, enqueue a **whole-layer warm** of layer `l+1`'s expert directory (prefill touches most experts of the next layer, so no routing prediction is needed — a plain sequential sweep of the bins is the right prefetch).
- Residency-gate per expert via the existing `expert_cold()` probe so hot bins are never re-read.
- Bound the read-ahead to one layer (double-buffer semantics) so the page cache isn't churned past what the pin budget protects.

Prefill is the TTFT term, and TTFT is what agent frameworks time out on — FreeToken's tail-TTFT comparison (44 s worst-case vs 232–946 s for llama.cpp/KTransformers-class baselines) is the motivating workload argument.

## Relation

- Complements #39 (multi-token sparse FFN prefill kernel) — that cuts prefill compute, this hides prefill I/O.
- The decode-side sibling of this idea (hot/cold overlapped expert reads) is in PR form separately.

Credit: opportunity identified while reviewing **FreeToken, arXiv:2608.16157** against the Cascadia codebase.

PR #145, opened 2026-08-22
perf(glm): hot/cold overlapped expert reads — FreeToken-style bandwidth split for NVMe-streaming decode

t8:
## Motivation

On the NVMe-streaming MoE models (GLM-5.2, DeepSeek-V4-class), decode tok/s is decided by how expert weights reach the FFN kernel. Today the per-token path is all-or-nothing:

- default: compute every routed expert from its mmap, faulting cold pages in mid-GEMV (random-read bound);
- `CASCADIA_GLM5_R1READ=1`: read **every** routed bin up-front in parallel, then compute (~1.4× on NVMe) — but it re-reads bins that are already page-cache resident (a wasted whole-bin copy per hot expert), and all reads complete **before** any compute starts, so read time and compute time add.

**FreeToken** (*Efficient Edge-Native MoE Serving with Bandwidth-Adaptive Execution*, Yang et al., [arXiv:2608.16157](https://arxiv.org/abs/2608.16157)) — the paper this design credit belongs to — divides each token's expert cache-misses between a transfer branch and a host-compute branch and runs the two **concurrently**, exposing `max(branches)` instead of their sum. On our UMA fleet there is no PCIe pool, so the two bandwidths are **page-cache-resident compute** vs **NVMe reads** — but the split-and-overlap structure carries over directly.

## What this PR does

New opt-in decode path, `CASCADIA_GLM5_HOTCOLD` (`crates/cascadia-engine-sparse-moe/src/glm/moe.rs`):

1. Classify each routed expert with the existing working-set probe (`expert_cold()`: `mincore` / `QueryWorkingSetEx`, <50% resident → cold).
2. **Cold** bins: one background I/O thread each does a whole-bin sequential read (same read as R1, and it warms the page cache as a side effect). Plain `std::thread` on purpose — the reads are I/O-bound and must not occupy the rayon pool the GEMVs saturate.
3. **Hot** experts and the always-pinned **shared expert** compute from resident mmap pages while those reads are in flight.
4. Drain the reads, compute each cold expert from its buffer (`swiglu_from`, byte-identical source), then accumulate **in gate order, then shared** — the exact op order of every other branch, so output is bit-identical.

Degenerate cases: all-hot → plain mmap loop, no threads, no copies (zero-overhead steady state); a failed read falls back to the mmap kernel for that expert. `CASCADIA_GLM5_HOTCOLD=cold` forces every mmap expert cold — deterministic test coverage of the threaded path and an upper-bound read benchmark. When both are set, HOTCOLD supersedes R1READ.

Observability: `CASCADIA_GLM5_PROFILE` dump now prints a `hotcold cold=…%` line (fraction of routed slots read off-thread), next to the existing hit% / eff MB/s.

Composes with the existing warming machinery rather than replacing it: LOOKAHEAD still predicts and warms layer L+1 (reducing the cold count); HOTCOLD covers whatever arrives cold anyway.

## Validation

- **Parity (bit-identical output)**: two new test binaries mirroring `glm5_r1.rs` against the committed GLM-5.2 export fixture — `glm5_hotcold_probe` (`=1`, exercises classify + all-hot fast path) and `glm5_hotcold_forced` (`=cold`, exercises the full threaded read + overlap machinery). Both produce the same greedy tokens as the mmap reference. Env-parse unit test included.
- **No regressions**: full `cascadia-engine-sparse-moe` suite green (all test binaries pass, including `glm5_r1` and `glm5_expert_mmap`); `cargo clippy --all-targets` adds no new warnings in the touched files; `cargo fmt` clean.
- **Measured (miner, synthetic-scale — see the validation comment below for the full matrix)**: zero overhead at full residency (+1.5% vs baseline), **+6.9%** at hit≈98%, and **1.97× baseline / 3.7× R1READ** in the memory-pressured streaming regime (cgroup-capped page cache, same hit%). All 40 bench runs bit-identical; real-binary `/v1/completions` e2e identical baseline vs HOTCOLD. Real-GLM-5.2 numbers still pending an export being staged.

## Follow-ups (out of scope here)

- DSV4's decode path shares `expert_mmap` but has its own forward; port once GLM numbers confirm.
- An adaptive q⋆-style refinement (choose how many cold experts to read concurrently from the measured NVMe MB/s vs GEMV throughput the profiler already collects) — plus the sibling opportunities from the same paper review: #141 (prefill next-layer streaming), #142 (semantic-anchored prefix checkpoints), #143 (recency-adaptive pins), #144 (elastic budget rebalance).

Credit: design follows **FreeToken, arXiv:2608.16157** (Yang, Fan, Pan, Xi, Wang, Sun, Keutzer, Han, Zaharia, Xu, Stoica).


t8:
## Hardware validation (miner, 2026-08-22)

Ran the fleet perf validation + e2e on the miner (Xeon Gold 6252, 48T, AVX-512, 172 GB RAM, WD SN770 NVMe, Ubuntu). No GLM-5.2 model artifacts exist on the fleet anymore (external SSD is full of K2.6), so the runs use a **22.7 GB synthetic GLM-5.2-format export** generated on the NVMe — same manifest/int4_bin contract as `export_glm5.py` (24 MoE layers × 96 routed experts + shared, 9.73 MB bins, top-8, H=4096, MI=1408), random-but-bounded weights. This exercises the identical loader, kernels, and I/O path; only the weight *values* are synthetic.

### Setup

- Harness: `glm5_run --force` (new commit on this branch) — teacher-forced decode of a fixed 128-id stream through the per-token path, because greedy generation on a synthetic model collapses to a fixed point (frozen routing → ~100% cache hit, nothing to measure).
- Residency regime controlled two ways: `posix_fadvise(DONTNEED)` eviction of the export before every run, and the run wrapped in a cgroup scope with `MemoryMax` ∈ {14G, 6G, 4G, 3G} (`MemorySwapMax=0`).
- `CASCADIA_GLM5_EXPERTS=mmap CASCADIA_GLM5_NOPIN=1 CASCADIA_GLM5_PROFILE=1`, 3 repeats per cell.

### Decode throughput (tok/s, mean of 3; profiler hit% in parens)

| MemoryMax | baseline mmap | R1READ | **HOTCOLD** | HOTCOLD vs base | vs R1 |
|---|---|---|---|---|---|
| 14G / 6G (working set fits) | 5.98 (98.6%) | 4.10 (98.6%) | **6.07** (98.6%, cold 1.5%) | +1.5% | +48% |
| 4G | 5.63 (97.9%) | 4.10 (98.5%) | **6.02** (98.6%, cold 1.5%) | +6.9% | +47% |
| 3G (pressured) | 2.54 (92.5%) | 1.36 (**11.0%**) | **5.00** (92.9%, cold 8.0%) | **+97% (1.97×)** | **3.7×** |

Diagnostic `HOTCOLD=cold` (force-all-cold upper bound): 4.33 tok/s at 4G, 2.38 at 3G — the all-read path alone beats R1 everywhere.

### What the numbers say

1. **Zero-overhead steady state confirmed**: at ~99% residency HOTCOLD ≡ baseline (+1.5%); the all-hot fast path adds no threads or copies.
2. **The pressured regime is the payoff**: at MemoryMax=3G, HOTCOLD is **1.97× baseline at the same hit%**. The win is exactly the PR's mechanism: baseline pays its ~8% misses as random 4K page faults *inside* the GEMV; HOTCOLD converts them to sequential whole-bin background reads hidden behind the hot experts' compute.
3. **R1's failure mode, measured**: under memory pressure R1's read-everything strategy evicts its own future hits — its hit rate collapses to 11% and throughput halves vs baseline. HOTCOLD reads only the misses and preserves the hot set (FreeToken's "don't spend bandwidth on hits", arXiv:2608.16157, empirically confirmed).

### Correctness / e2e

- **Bit-parity**: all 40 bench runs across configs/caps print the identical final-logits FNV (`0xe9f4635c4859a496` for the forced stream); the earlier greedy matrix (20 runs) produced identical 48-token outputs.
- **Full crate suite green on the miner** (Linux + AVX-512 kernel paths): all 50 test binaries pass in release, including the two new parity tests.
- **Real binary e2e**: `cascadia worker --engine sparse-moe --model <synth> --api 127.0.0.1:8077` (manifest `arch=glm5` → GlmRunner), `/v1/completions` with `temperature: 0` — completions **identical** between baseline and `CASCADIA_GLM5_HOTCOLD=1`, worker logs clean, graceful shutdown.

### Caveats

- Synthetic weights concentrate routing (~370 of 2304 bins form the hot set), which is why the residency regime had to be swept with cgroup caps rather than emerging from model scale. Real-GLM-5.2 numbers (broader routing, bigger bins) remain worth collecting when an export is staged again; the mechanism-level result — fault-to-overlapped-read conversion at mid hit% — is what transfers.
- Bench artifacts kept on the miner for follow-ups (#141/#143): export at `~/glm5-synth`, logs in `~/hotcold-bench*/`, worktree `~/cascadia-hotcold`.

7i7o:
## Rebase + review + hardware re-validation (tip `b95ba16a`)

Rebased onto `main` (now through #151/#152/#153) — no conflicts — then ran a full parallel review (code / silent-failure / tests / comments / types). The review confirmed the core is sound (bit-identical output, sound `thread::scope` concurrency, shared-expert-computed-once, correct cleanup) and surfaced a set of robustness / observability / coverage gaps, all now fixed (one commit each):

**Robustness**
- `read_bytes` now returns `Err` on a short read (a bin truncated/replaced after `open` returned `Ok(short_vec)` that sliced OOB in `swiglu_from` — the "failed read falls back" promise didn't cover this; fixes both hot/cold and R1).
- Cold-read spawn uses `Builder::spawn_scoped`; a thread-creation failure under resource pressure now computes the slot inline/hot instead of panicking out of decode.
- Warn once when the residency probe (`mincore`/`QueryWorkingSetEx`) is unavailable, so a silently all-hot no-op is visible.

**Observability**
- Cold-read fallbacks (I/O error or panicked reader) are counted (`HC_COLD_FAIL`, surfaced as `fail=N` on the PROFILE `hotcold` line) and warned once — a 100%-failing overlap was previously indistinguishable from a working one behind bit-identical output.

**Coverage** (the bit-identity claim was only argmax-checked before)
- Unit test pinning `swiglu_from(read_bytes)` == `swiglu_mmap` **bit-for-bit**, plus the short-read → `Err` fallback.
- Both parity binaries now assert via the profiler counters that the intended path actually ran; new binary deterministically covers the all-hot `ncold==0` fast path.

**Misc**: `glm5_run --force/--ids` arg guard + usage + empty-logits assert; corrected comments (overlap covers reads not cold compute; "probe residency" is Probe-only; read computes from the owned buffer).

### CI (local, all green)
`cargo fmt --all --check` · `clippy --workspace --all-targets` · `build -p cascadia` · `test --workspace --all-targets` · kv-coord `test` + `check`.

### Hardware re-validation (miner — Xeon Gold 6252, AVX-512, WD SN770 NVMe, Ubuntu)

Same harness as the 2026-08-22 run: deterministic (seed 7) 22.7 GB synthetic GLM-5.2 export, `glm5_run --force` teacher-forced decode of a fixed 128-id stream, export `fadvise(DONTNEED)`-evicted before each run, `MemoryMax` cgroup caps, `CASCADIA_GLM5_EXPERTS=mmap NOPIN=1 PROFILE=1`, 2 reps/cell.

Decode tok/s (mean of 2; profiler hit% / hotcold cold%):

| MemoryMax | baseline | R1READ | **HOTCOLD** | vs base | vs R1 |
|---|---|---|---|---|---|
| 14G (fits) | 5.61 (98.6%) | 3.82 (98.6%) | **5.81** (98.6%, cold 1.5%) | +3.6% | +52% |
| 4G | 5.39 (97.9%) | 3.82 (98.5%) | **5.74** (98.6%, cold 1.5%) | +6.5% | +50% |
| 3G (pressured) | 2.52 (93%) | 1.36 (**11.0%**) | **4.95** (93%, cold 7.7%) | **+96% (1.96×)** | **3.6×** |

Diagnostic `HOTCOLD=cold` at 3G (force-all-cold): 2.43 tok/s, `cold=100.0%`.

- **Bit-parity**: all 19 forced runs (every config × cap × rep) print the identical `argmax=7874 logits_fnv=0xe9f4635c4859a496` — the same FNV as the 2026-08-22 run, so the review fixes changed no numerics.
- **R1's failure mode reproduced**: at 3G its hit rate collapses to 11% (reads evict its own future hits) and it runs at half of baseline; HOTCOLD reads only the misses and holds 93% hit.
- **Full `cascadia-engine-sparse-moe` suite green on the miner in release** (Linux + AVX-512): 55 test binaries, 0 failed — including the new bit-identity/short-read/all-hot tests, so `swiglu_from` == the mmap kernel is confirmed on the real AVX-512 GEMV path.
- **Real-binary e2e**: `cascadia worker --engine sparse-moe`, `/v1/completions` (`temperature 0`) — completions **identical** baseline vs `HOTCOLD=1` (`t7874 t2726 …`), worker log shows `hotcold cold=1.1% (11/960 routed slots read off-thread)` with no `fail=`, graceful SIGTERM shutdown.

Same caveat as before: synthetic weights (routing concentrated, residency swept via cgroup caps); real-GLM-5.2 numbers still worth collecting when an export is staged. The mechanism-level result — fault-to-overlapped-read conversion at mid hit% — transfers.

7i7o (approved):
tACK 👍

PR #146, opened 2026-08-22
perf(glm): stream next layer's expert set during batched prefill — FreeToken-style prefill overlap

### LINKED ISSUE #141: Stream layer L+1's expert set during batched prefill — overlap NVMe reads with compute

t8:
Closes #141. **Stacked on #145** (`perf/hot-cold-expert-overlap`) — review only the top two commits; the base commits are #145's.

## What this does

New opt-in `CASCADIA_GLM5_PREFILL_STREAM`: while the batch-union prefill computes layer `i`, the shared lookahead worker sequentially warms layer `i+1`'s not-yet-resident routed expert bins into the OS page cache (and layer 0's before the loop, so its warm overlaps layer 0's per-row attention). Prefill's expert reads then overlap compute instead of serializing between layers — FreeToken's full-layer double-buffered prefill (*Yang et al., arXiv:2608.16157*, the paper credited for this opportunity) mapped onto the page cache.

Design points:

- **No routing prediction needed.** A wide batch's routed union covers ~the whole next layer (64 rows × top-8 draws ≈ full set), so the whole cold set is warmed, gated per expert by the existing working-set residency probe.
- **Row threshold.** A narrow batch's union is far smaller, and whole-layer warming over-reads — measured 1,552 bins warmed on a 16-row prompt whose union needed a few hundred. Gated streaming stands down below 64 rows; `=all` skips both the gate and the threshold (deterministic test coverage + bench lever).
- **Shared worker, decoupled flags.** The `glm5-lookahead` worker is now spawned for either consumer; decode-side router-proxy lookahead stays gated on `CASCADIA_GLM5_LOOKAHEAD` alone (new `lookahead_decode` field).
- **Output-inert.** Warming is enqueue-only, best-effort page-cache traffic. Parity test binary included (`glm5_prefill_stream`, `=all` on the committed fixture).

## Measured (miner, synthetic-scale — same methodology as #145's validation)

Cold batched prefill of a 192-id prompt, export fadvise-evicted before every run, 3 reps. Synthetic GLM-5.2-format export (§caveats in the #145 comment apply).

**NVMe (WD SN770, ~gigabytes/s) — honest null result:** base 26.2–27.1 s, streamed 27.1–27.4 s. Warm-cache floor is 25.3 s, i.e. batched prefill on local NVMe at CPU compute speeds is ~95% compute-bound — the 48-thread mmap fault path already overlaps nearly all I/O, so there is almost nothing to hide and the worker's extra reads cost ~1%. FreeToken's 26% overlap win presumes *transfer-bound* prefill (PCIe + GPU compute); that balance does not exist on this box.

**USB-attached SSD (SanDisk Extreme Pro, ~1 GB/s — the fleet's "model dir on external storage" case), 13-layer export:** streamed wins every paired rep — base 18.5/19.9/26.4 s vs streamed 16.5/17.6/19.2 s, mean **21.6 → 17.8 s (−18% TTFT, up to −27%)**, and far more run-to-run stable. Warm-cache floor 12.7 s → streaming hid ~43% of the exposed I/O (8.9 s → 5.1 s). All 1,152 warms landed, zero stale drops.

**Short-prompt guard:** 16-row prompts: streamed ≡ base (3.4 s, `warmed=0`) — the threshold stands down as designed.

**Correctness:** parity across every run (identical out ids); full crate suite green locally and on the miner (release, Linux/AVX-512); real-binary e2e — `cascadia worker --engine sparse-moe` + `/v1/completions` with a 96-token prompt — identical completions base vs streamed, worker log shows the streamer engaged, no errors.

## When to enable

Model dir on slow storage (external SSD, network mount): measurable TTFT win. Local NVMe with CPU experts: neutral — leave off (the default). The knob should get more valuable wherever per-byte compute gets faster relative to storage (iGPU-offloaded experts, wider batches).

## Follow-ups

- A small reader pool (2–3 warm threads) could reclaim more of the remaining exposed I/O on slow storage (single reader hid ~43%).
- DSV4 shares the staged prefill driver but returns `supports_batched_prefill() = false`; nothing to do there until it batches.

Credit: **FreeToken, arXiv:2608.16157** (Yang, Fan, Pan, Xi, Wang, Sun, Keutzer, Han, Zaharia, Xu, Stoica).


7i7o:
## Rebase + full review + fixes + hardware re-validation

Rebased onto `main` (which now carries #145) with `--onto`, dropping the two **pre-rebase #145 commits** this branch still carried (`419ac053` hot/cold, `c9bcecc5` `--force`); a diff-of-diffs confirmed the two real #146 commits replayed byte-identically (no content lost). Branch is now `main` + exactly the two #146 commits + the fixes below.

### Review
Five parallel agents (code / silent-failure / tests / comments / types). **No correctness blocker** — the change is output-preserving; the decode-guard refactor (`lookahead_decode`) is behavior-preserving; the pre-loop/loop double-buffering enqueues each layer once; edge cases (`n==1`, dense layers, `rows==0`) are safe. Findings were a misplaced doc, test-coverage gaps, and observability gaps around the `warmed=N` claim.

### Fixes (one commit each)
1. `docs`: restore `save_usage`'s doc (it had merged into `stream_enqueue_layer`'s `///` block).
2. `test`: extract pure `prefill_stream_gate(mode, rows)` + table-test the row threshold (`<64` stand-down, `All` exempt, boundary).
3. `test`: decode-lookahead output-parity smoke test (`CASCADIA_GLM5_LOOKAHEAD=1` + mmap) — that guarded path had no net.
4. `feat`: count prefill-stream enqueues + assert streaming actually engaged in the parity test (output-parity alone can't catch a silent no-op).
5. `feat`: log the resolved mode (`off|gated|all`) at load — an `all` A/B lever typo silently became `gated` before.
6. `feat`: warn when the residency probe is dead on the gated prefill path (extracted `MoeLayer::residency_probe_dead`, reused by the hot/cold warning).
7. `feat`: note stand-down for the other silent `warmed=0` causes (non-mmap rank, narrow batch).
8. `docs`: correct comments — conditional prefetch precedence, hot-bin re-read is RAM not NVMe, drop the one-off `1552`/`64×8` bench figures.

### CI — all 4 lanes green
`cargo test (stub)` (fmt/clippy/build/workspace-test), `cargo test (kv_coord engines)`, `cargo-deny`, `build + test (openvino feature, ai-pc)`. Also re-run green on the miner (Linux/AVX-512): 100 test-result lines, 0 failures, incl. the 4 new tests.

### Hardware re-validation (miner, seed-7 22.7 GB synthetic export regenerated, removed after)

**Cold batched prefill, 192-id prompt, 3 reps (NVMe — honest null result, as claimed):**

| cfg | rep1 | rep2 | rep3 | warmed/dropped |
|-----|------|------|------|----------------|
| base | 26.3 | 27.0 | 27.0 | — |
| gated | 27.3 | 27.1 | 26.9 | 2304 / 0 |
| all | 26.9 | 26.9 | 27.0 | 2304 / 0 |

Base ~26.8s vs streamed ~27.0s — neutral (~+0.7%); batched prefill on local NVMe is compute-bound, nothing to hide. Matches the PR's `base 26.2–27.1s / streamed 27.1–27.4s`.

**Short-prompt guard (16-id):** base ≡ gated (3.4s, `warmed=0`, stands down), out ids identical. New diagnostic fires: `gated streaming stands down below 64 rows (this batch: 16)`. `all` (threshold-exempt) warmed 980 / dropped 1310 and ran slower (4.7s) — exactly the narrow-batch over-read the threshold exists to avoid.

**Correctness:** every run (192-id ×9, 16-id ×6) produced out ids `[7874]` — bit-identical base vs streamed. Real-binary e2e (`cascadia worker --engine sparse-moe` + `/v1/completions`, 96-token prompt): base ≡ gated ≡ all completions identical, worker log shows the streamer engaged (`prefill_stream=gated warmed=2112 dropped=0`), no errors, graceful shutdown.

**Not reproduced:** the USB-attached-SSD result (−18% TTFT) needs external slow storage; the miner has only NVMe, so that specific number can't be exercised here — but it is the case where the null result above predicts a win.

7i7o (approved):
tACK 👍

PR #147, opened 2026-08-24
ci: add Linux openvino check for shim and engine

twilson63:
Pre-flight the release compile that only runs at tag time (`cargo check --features openvino` on `ubuntu-22.04` + `g++-12` + OV SDK).

**Why:** stub `ci` never built the C++ shim, so the GCC bool/int lambda error in `gemv_offload.cpp:143` passed CI and broke `v0.1.3`/`v0.1.4` mid-publish (linux bundle missing, release stuck `prerelease`). This mirrors `release.yml: build-linux`.

**Gated:** via `dorny/paths-filter` — only runs when `crates/cascadia-ov-genai-shim`, `crates/cascadia-engine-openvino`, or `crates/cascadia-engine-sparse-moe/examples` change. Zero extra cycles on unrelated PRs, always on `push:main`/`workflow_dispatch`.

Follows repo conventions: SHA-pinned actions, `persist-credentials: false`, `contents: read`, `OV_LINUX_URL` as workflow `env`, `timeout-minutes`.


7i7o (approved):
tACK 👍

PR #151, opened 2026-09-03
fix(gemma4-export): mask sliding_attention layers to their window

t8:
Gemma 4 alternates five `sliding_attention` layers with one `full_attention` layer, but
`export_gemma4.py` built a plain causal mask for every layer, so sliding layers attended to the
whole prefix. That matches HF only while the sequence stays shorter than the window (1024 on
31B-it): short prompts look correct and long ones silently degrade, which is why this survived.

The existing "Faithful sliding-window masking before long-context use" note in the enterprise
gemma4 runbook records the gap; this closes the exporter half of it.

## Evidence

Observed on a 3-stage int4 31B-it deployment across three Arc 140V nodes. A 2,951-token page came
back generic and then degenerate after five bubbles; after re-exporting with this change the same
page is faithful on all five, and a 3,386-token page stays coherent and well-formed across nine.
The `.bin` files differ by exactly 8 bytes between the two exports (the window constant); the
weights are unchanged.

## What changed

- `build_attention_mask()` is extracted and shared by the dense and KV-shared layer-forward paths,
  instead of the causal-mask block being duplicated in both. It raises `ValueError` for a window
  `<= 0`, which would otherwise mask every key and produce NaN softmax silently.
- Sliding layers get the window band; `full_attention` layers pass `None` and are unaffected.
- `resolve_sliding_window(text_config)` is the single source of truth: 0 or absent disables the
  bound (so architectures without a window export exactly as before), negative is a configuration
  error.
- `stage_config.json` and `pipeline_config.json` record `"sliding_window"`, and `export_version`
  is bumped to `gemma4_cached_v1.1` so pre-fix exports are distinguishable on disk. Safe because
  every reader matches by prefix: `genai.rs:165` (`starts_with("gemma4")`), `dist_spec.rs`
  (`starts_with("v3")` rejections only), `runtime.rs:71,89` (deserialised, never compared), and
  no `deny_unknown_fields` anywhere under `crates/`.

## Testing

`tools/tests/test_gemma4_sliding_window.py`, 29 tests:
- the mask contract: key `j` is visible to query `i` exactly when `0 <= (past + i) - j < window`,
  over prefill, decode against a cache, chunked prefill, window equal to and wider than the cache,
  window of one, the strict-subset and no-fully-masked-row invariants, and the `<= 0` guard;
- the resolver: 1024, 512, 0, null, absent, negative;
- the production call sites: a small fake decoder layer driven through
  `cached_gemma4_layer_forward` and `cached_gemma4_shared_layer_forward`, asserting `window=4`
  differs from `None` once `past + seq > 4` (with identical returned KV) and is a no-op when the
  cache fits.

Full `tools/tests/` suite: 112 passed, 1 pre-existing skip. Independently, an adversarial review
swept 960 shape/window cases against HF's `sliding_window_overlay ∧ causal` rule with zero
mismatches, and traced the real layer forward through `ov.convert_model` (`aten::tril` with a
`ShapeOf`-derived diagonal lowers to `Range/LessEqual/Select`; the saved IR matches torch to 1e-6
on unseen shapes).

## Not covered

- The runtime half of the same gap is still open: the optimum-intel OpenVINO export of Gemma 4
  served through OpenVINO GenAI degrades on long context independently of this exporter,
  reproduced on both an Intel iGPU and a CPU-only host. Details in labscommunity/cascadia-enterprise#114.
- Pre-existing and out of scope: `test_gemma4_dispatch.py` calls `export_shards.main()`, which
  sets `torch.set_default_dtype(torch.float16)` and never restores it, so running it before
  `test_arch_detection.py` fails six tests there. A `set_default_dtype` fixture in the dispatch
  test would fix it. CI does not currently run `tools/tests/`.

## Review follow-ups (fd73474d)

A nine-angle review of this PR (one of them re-drove the whole `CachedStageWrapper` against HF `Gemma4ForCausalLM` at prefill/decode/chunked prefill past the window, incl. KV-shared and cross-stage external KV, and matched to ≤ 6.6e-7) found the mask correct and raised these, all addressed:

- **Windowless configs were stamped v1.1 while exporting the v1 behaviour.** `resolve_sliding_window` now takes `layer_types` and refuses a config with `sliding_attention` layers but no positive window (HF refuses that mask too); bool/non-integral values, per-layer overrides, bidirectional (embedding) variants and unknown layer types are rejected config-first, before the multi-minute model load (previously the negative-window check fired after it).
- **Pre-fix trees loaded silently.** The `gemma4` engine now reads `export_version` / `layer_types` / `sliding_window` from `stage_config.json` and warns at load when a sliding layer has no window, naming the stage dir and asking for a re-export.
- **One mask per layer in the traced graph.** The wrapper builds one additive mask per (window, cache source) per forward and passes it into both layer forwards; `build_attention_mask` shares a single `-inf` plane (verified graph-equal). Per-layer windows follow the file's per-layer-list pattern with `== "sliding_attention"` like HF.
- **Self-verify never crossed the window.** `_verify_stage` now prefills `sliding_window + 4` tokens so the traced `tril` band executes, and asserts finite outputs.
- **Stale `gemma4_cached_v1` strings** in `cascadia engines`, rustdoc, the engine module doc, README, CLI.md, ARCHITECTURE.md and the genai staged-tree test; `EXPORT_VERSION` constant; gemma4-support.md documents the window and the re-export requirement.
- **Tests** (45): HF `sliding_window_causal_mask_function` cross-check, the config-first resolver contract, an exact oracle for the layer forwards (windowed step == full attention over the truncated cache; one key wider fails), a shared-layer decode case, and a tiny random-init `Gemma4ForCausalLM` driven through `build_cached_wrapper` against HF's own logits past the window (skips when transformers lacks Gemma 4).

Validation caveat surfaced by the review (pre-existing, not changed here): on transformers ≥ 5.5 a heterogeneous config (31B's `head_dim=256` / `global_head_dim=512`) raises `AmbiguousGlobalPerLayerAttributeError` on the exporter's global `text_config.head_dim` reads, so 31B exports on that transformers need `allow_global_per_layer_attribute_access` or the bundled remote code.


7i7o (approved):
tACK 👍

PR #152, opened 2026-09-04
feat(qwen35): dense Qwen3.8-27B (qwen3_5) through the Qwen3.6 surgery + staged engine

t8:
## Summary

Serve dense **Qwen3.8-27B-class** models (`model_type: qwen3_5`, hybrid Gated DeltaNet + attention) through the machinery that already serves Qwen3.5/3.6 MoE (`qwen3_5_moe`). Full write-up: `docs/architectures/qwen3.8.md`.

- `tools/qwen36_surgery/export_qwen36_moe.py` is config-driven: hidden size, layer count and per-layer attention type come from `config.json`; state-variable ids walk `layer_types`; stage splits that would leave a stage without a full-attention layer are refused (≤16 stages for the 64-layer 27B); the manifest records `family` / `hidden_size` / `num_layers` / `layer_types`.
- `qwen36.rs`: `hidden_size` from the manifest (default 2048 for Qwen3.6-era trees) sizes the activation frames; `arch` may be `qwen3_5` or `qwen3_5_moe`.
- CLI: `--engine qwen35` (`qwen36-moe` kept as an alias); `cascadia shard` dispatches any `qwen3_5*` IR dir to the surgery; post-shard hint updated.
- `export_shards.py`: dense `qwen3_5` safetensors are hard-rejected with a pointer to the IR route (`"qwen3_5"` contains `"qwen3"` and would otherwise mis-export through the Qwen3 dense path); `linear_attention` layer types are a hard quirk; `*.jinja` rides along with downloaded IRs.
- Aliases `qwen3.8-27b` → `OpenVINO/Qwen3.8-27B-int4-ov`, `qwen3.6-35b-ov`.
- Tests: surgery shape helpers, `qwen3_5` dispatch, arch-detection cases; the parity gate is parametrised (`QWEN36_SHARDS` / `QWEN38_SHARDS`).

CI: fmt, clippy, stub + kv_coord + openvino-feature AI-PC jobs green.

## Validation — tate-07 (Core Ultra X7 358H, Arc B390 iGPU, 64 GB LPDDR5X)

Model: `Jackrong/Qwopus3.8-27B-Flash` (Qwen3.8-27B fine-tune), exported to int4 with the recipe in the doc (Intel's INT4_ASYM g128; byte-size-identical layout to `OpenVINO/Qwen3.8-27B-int4-ov`).

**Correctness.** HF `transformers 5.2.0` bf16 greedy reference (Mac Pro, 1.5 TB RAM) vs the OpenVINO exports, 32 greedy tokens: the **bf16 IR matches the reference token-for-token** on every prompt whose first step isn't an exact bf16 logit tie; the int4 IR matches on both raw prompts and `17×23 = 391`, and differs only at quantisation near-ties (a spacing token in the primes list, the tied haiku). 2-stage surgery `--validate`: top-1 match, top-5 5/5, 8/8 greedy chain-vs-whole. `cascadia shard` (real entry point) on the IR dir: dispatched + validated, exit 0. Every answer correct on CPU and GPU, identical text between devices.

**Throughput through `cascadia run … --api` → `/v1/chat/completions`** (19-token prompt, 96-token completion, greedy; `tok/s` = completion tokens / whole-request wall time, so prefill and HTTP are included):

| path | SDK | device | load | 96-token request | tok/s |
|---|---|---|---|---|---|
| `--engine qwen35`, 2-stage chain | GenAI 2026.2.1 (repo pin) | B390 iGPU | 41 s | 15.0–15.6 s | **6.2–6.4** |
| `--engine qwen35`, 2-stage chain | GenAI 2026.2.1 | CPU | ~80 s | 32.2 s | 3.0 |
| `--engine ov-genai`, whole IR | GenAI 2026.3.0 | B390 iGPU | ~60 s | 13.2 s | **7.3** |

Raw-IR decode (no HTTP, 5-token prompt): B390 6.4–6.7 tok/s, CPU 3.8 tok/s; Intel's own IR 6.3–6.4 / 3.7. Short-prompt TTFT on the B390 is 0.2–0.7 s (5–30-token prompts, warm); the first inference of a fresh process on CPU pays a 20–30 s one-off JIT warm-up.

**Context capacity and TTFT** (Qwopus int4, B390, weights resident, synthetic prompt, 512-token prefill chunks, 16 greedy tokens after; fresh process per point; resident = host-visible RSS including the iGPU's UMA allocations):

| prompt tokens | TTFT | prefill tok/s | decode tok/s | resident |
|---|---|---|---|---|
| 1 K | 2.4 s | 425 | 6.1 | 17.1 GB |
| 4 K | 8.1 s | 505 | 6.4 | 18.1 GB |
| 8 K | 15.5 s | 527 | 6.1 | 19.1 GB |
| 16 K | 32.8 s | 500 | 6.2 | 19.5 GB |
| 32 K | 78.5 s | 417 | 5.1 | 20.8 GB |
| 64 K | 219 s | 299 | 4.0 | 21.9 GB |
| 128 K | 667 s | 196 | 3.1 | 25.1 GB |
| 256 K (native max) | 3256 s (54 min) | 81 | 2.5 | 31.9 GB at decode, ~43 GB peak in prefill |

Memory is not the limit on 64 GB (state grows ~60–120 KB/token: 64 KB/token of KV over the 16 attention layers, fixed-size DeltaNet state on the other 48); time is. ≤32 K is interactive, 64–128 K is usable with minutes of TTFT, 256 K is a capacity fact rather than a setting on an iGPU.

**Prefix caching (new in this PR, `qwen35`, always on).** `crates/cascadia-engine-openvino/src/prefix_cache.rs`: a byte-bounded LRU of per-stage chain-state snapshots keyed by exact token prefix; admission restores the longest cached strict prefix over the live requests and prefills only the tail. Snapshot positions are the **end of the system block** (before the second `<|im_start|>`, so a new conversation on the same system prompt starts warm) and the **chat boundary** (before the prompt's last `<|im_start|>` — the family's template renders history assistant turns without the `<think>` block the live generation prompt carries, so a snapshot keyed on the previous turn's full sequence would never match; the prefill span is split there). Non-consuming (a shared system prompt serves every conversation). Restores prime un-run requests with one dummy fold first (the GPU plugin silently drops a `set_state` onto a request that has not executed since its reset; onto a live one it holds bit-exactly, measured Δ logits 0.0), in the order clear → prime → restore. A cold turn snapshots every boundary; a warm turn **refreshes** the boundary snapshot only once the tail it would save has grown to cost as much as the copy, with both rates measured by the engine itself (`RefreshCosts`: prefill tok/s, seconds per byte copied), so short follow-ups pay restore + tail only, the tail never exceeds one copy's worth of prefill, and a conversation never goes cold. Snapshots are refused while a stage reset has failed (`stages_dirty`), and `cancel()` resets. `--prefix-cache-gb` (default min(16 GiB, RAM/4); 0 = off) on `run`/`worker`. Single-process only; pipeline mode keeps the `kv_coord` plane's CAPTURE/RESTORE for multi-rank. The plane's blob framing helpers moved into the new module (re-exported); the in-process cache keeps parts unframed (no u32 length bound, no copy on hit).

**Certified** (tate-07, B390, 2-stage `qwen35` through `cascadia run --api`; the same 4-turn script against a cache-on and a cache-off server, greedy; every warm answer identical to cold):

| first turn | follow-up TTFT (cache off → on) | new conversation, same system prompt (off → on) | cold turn cost (two snapshots) |
|---|---|---|---|
| 8 K tokens | 16.9 s → **2.1 s** (7.9×) | 15.8 s → **1.9 s** (8.5×) | 17.8 s → 23.4 s |
| 32 K tokens | 91.5 s → **5.5 s** (16.6×) | 89.9 s → **5.8 s** (15.4×) | 92.9 s → 111.3 s |

Restore is a host→device copy (1.4 s for 1.2 GB at 8 K, 4.3–5.1 s for 4.45 GB at 32 K); a snapshot costs ~1.8× that. The eviction rule was exercised separately (8 K, 3 GiB budget, a 2,500-token turn): the refresh evicted the old boundary and the new conversation on the same system prompt still started warm (1.9 s TTFT). Full per-turn tables in the doc.

**Other guards from review.** Thinking requests against Qwen3.8 no longer 400: the API's GLM-mapped `reasoning_effort` ("high") is rejected by this family's template (`xhigh/medium/low`), so `render_or_fallback` retries once: with the caller's own word when they sent one that differs from the mapping, otherwise with the effort undefined (template default); a caller word the template also rejects is a 400 (changed in the 2026-09-08 round); verified on hardware (default, low, medium, xhigh, thinking-off all 200). `--api-max-body-mb` defaults to 1 MiB (~250 K tokens) and is validated; the engine additionally refuses prompts at or past `max_position_embeddings` at submit (413) rather than prefilling for an hour on the batch=1 slot. `cascadia shard --target npu` is rejected config-first for this family.

**Intel's published IR + `ov-genai`.** Its tokenizer IRs were built with openvino-tokenizers 2026.4 nightly (stateful ops); GenAI 2026.2/2026.3 throw in `VLMPipeline`. Regenerating them with the installed `convert_tokenizer --with-detokenizer` makes it serve; documented in the page. Exports made with the 2026.3 toolchain need nothing.

**Regression (Qwen3.6-35B-A3B).** Same exporter + engine on `OpenVINO/Qwen3.6-35B-A3B-int4-ov`: identical cut (40 states/stage, top-1 match, top-5 5/5, 8/8 chain-vs-whole), served via `--engine qwen35` on the B390 at 19–20 tok/s with correct answers; a manifest stripped to the Qwen3.6-era keys (no `hidden_size`) serves identically through the `--engine qwen36-moe` alias.

Not covered: NPU (out of scope for a 14 GB dynamic-shape IR); MTP-head speculative decoding (exported, unused; follow-up); vision input.

**Review round (2026-09-06).** Findings resolved in `16bdce0a`..`471c8063`: thinking requests 400'd on this family's `reasoning_effort` vocabulary; u32 blob framing bounded cached parts at 4 GiB; 4 MiB body cap with no token guard; the `MAX_WARM_TAIL` cut-over made long-answer conversations cold; unclamped 16 GiB cache default; unhittable turn-end snapshot delaying `[DONE]`; `cancel()` without reset; priming before the restore-clear A/B arm; `primed` set before the fold ran; failed reset could poison the cache; flag validation; `--target npu` on the surgery path; test dtype leak; doc headline recipe. Two more found by re-certification: the every-turn snapshot policy's copy inside each warm turn (→ measured refresh rule) and the shared system block losing the LRU race to its own conversations' refreshes (→ every extended prefix is marked used).

**Review round (2026-09-08).** Findings resolved in `4e85c1d9`..`62d8daa6`: the snapshot copy was paid before the budget check (now an exact refused-length threshold in `PrefixCache`, so an unservable budget wastes one copy per size class per process instead of one per turn); a caller's own rejected `reasoning_effort` word is a 400, only the API's mapped word falls to the template default; a failed stage reset refuses the turn with an error chunk and escalates to `recreate_request` instead of serving on unknown state; the 413 guard counts the injected `<think>` block and resume ids, `max_tokens` is capped at the window, and the final chunk carries `finish_reason` (`length` at the window); `hidden_size` is required for `qwen3_5` manifests (2048 only for `qwen3_5_moe`) with embedding-width and stage-order checks at load; `--prefix-cache-gb` is bounded by physical RAM (the saturating cast was unbounded) and both new flags are validated before the model loads and warned when ignored; `cascadia shard --validate` exits non-zero on a parity fail; a wrong-length `layer_types` list is rejected and the manifest `arch` is the family type even under a wrapper `model_type`; missing `<|im_start|>` / unknown `max_position_embeddings` warn at load; snapshot sizes corrected everywhere to the measured ~130 KB/token (1.2 GB at 8 K, 4.45 GB at 32 K, ~17 GB at 128 K); the cache is documented as process-wide and tenant-blind; stale comments from the superseded snapshot policies removed.

**Second pass (2026-09-08, tate-07, tip `62d8daa6` after the review round).** Same box, trees, scripts and budgets, rebuilt from the tip; full tables under "Second pass" in `docs/architectures/qwen3.8.md`. Functional probes (correct answers, `finish_reason` stop/length, streaming `[DONE]`, 413 body cap, thinking on/off and every effort word 200); throughput `qwen35` 2-stage B390 **6.6 tok/s** (was 6.2–6.4), CPU 2.8–2.9, `ov-genai` whole IR **7.4–7.6** (was 7.3); context sweep 1 K–256 K within noise of the first pass (256 K TTFT 47 min, 31.9 GB at decode); prefix cache 8 K follow-up TTFT 17.7 s → **2.2 s**, 32 K 91.2 s → **5.8 s**, new conversation on the shared system prompt 1.9 s / 5.8 s, eviction run reproduced, every warm text equal to its cold text; both parity goldens verify through the `enable_thinking = true` harness; Qwen3.6-35B regression 20–21 tok/s through `qwen35` and the legacy alias. The per-turn `recreate_request()` on warm turns measures 0.6–0.9 ms (`reset_s` on the task-done line), so it stays.


7i7o (approved):
tACK 👍

PR #154, opened 2026-09-11
feat(sparse-moe): Inkling (Thinking Machines 975B MoE) family

t8:
## Summary

Adds Thinking Machines' **Inkling** (975B / 41B active; Inkling-Small 276B / 12B is the same architecture at smaller dims) as a new pure-Rust family of the sparse-MoE engine, beside GLM-5 and DeepSeek-V4. Text-only, greedy/sampled decode, N-rank pipeline through the existing `StagedRunner` / `PipelineEngine`. The full 975B checkpoint has been exported, validated layer-for-layer on real weights, and served end to end through `cascadia run --api` on the miner. Details and every number: `docs/architectures/inkling.md`.

What is new versus the existing families (semantics from transformers' native `modeling_inkling.py`, which is also the test oracle):
- **No RoPE**: a learned relative-position bias (`r_proj · proj` bank, extent 1024 global / 512 sliding) with log-scaled queries on global layers.
- **Causal short convolutions** (kernel 4, residual inside) on k, v, the attention output and the MLP output — four per layer, each with decode state, rewind and snapshot/restore.
- **GQA** with 55 sliding-window-512 layers (512 + 32 row rings) and 11 global layers; per-head q/k RMSNorm; `1/D` scaling.
- **Router** normalising over the selected **and** the two shared experts (each shared expert gets its own gamma); `mlp.gate.bias` is selection-only.
- Embedding RMSNorm; mup-scaled logits sliced to the unpadded vocab.

`tools/export_inkling.py` streams the 1.905 TB checkpoint through int4 group-32 quantisation on a disk that cannot hold it: every tensor is staged the moment its shard is readable (a layer's tensors span shards 1..108, so per-unit consumption would have needed ~1.5 TB resident), units assemble when their parts exist, shards are deleted once consumed. On the miner (48 threads, SATA scratch) the whole export took ~2 h at ~350 MB/s download, ~1.2 s per expert; 512 GB on disk.

Two API defects surfaced by the real run, fixed here: `tojson` now honours `sort_keys` / `separators` (Inkling's template uses both for tool specs — tool requests would have 400'd), and the caller's `reasoning_effort` word reaches templates with their own scale (the GLM mapping had turned `"none"` into `"high"`, so thinking could not be switched off).

## Validation
- **Synthetic (CI):** primitive goldens vs HF 1e-4; tiny-model hidden states ≤ 0.1 % of row scale, 8/8 greedy ids exact; loader round-trip token-exact through int4; 2-rank chain over the real transport == single process; 31 pytest cases (exporter contract, de-interleave vs transformers' `Interleave`, late-arriving-shard streaming byte-identical to a plain export, direct-write vs staged identity, parity harness on the tiny export). The fixtures' relative-position bias is load-bearing (rms|bias|/rms|q·k/D| 1.3–2.3 per layer; zeroing it moves hidden states 4–9 % of row scale against a 2 % tolerance).
- **Real weights (975B export, layers 0–2 vs transformers on the same int4-dequantised weights, float32):** rms(diff)/rms 0.31 / 0.16 / 0.13 %, max|diff| ≤ 0.5 % of each row's max, cosine 1.000000, decode == prefill bit-for-bit. Yardstick: the model's own **bfloat16** execution deviates from float32 by 3.1 % rms on the MoE layer, ~25× more than the Rust shell.
- **End to end (real binary + API, greedy):** thinking off → `Paris`, `42`, `The Pacific Ocean is the largest on Earth, covering about 63 million square…`; thinking on (`reasoning_effort: "low"`) → `<think>The user asks for the capital of Italy, requesting one word. The answer is Rome.</think>Rome` (the special-token framing translated by the API). Server load 60–91 s; cold TTFT 107–176 s for 25 tokens depending on page-cache state; decode 8–25 s/token (0.05–0.13 tok/s) with the 490 GB of routed experts paging from a SATA SSD into 172 GB of RAM.

## Review round (2026-09-11)
`/code-review` on this PR returned 15 findings; 14 are resolved in `f914d65f`..`a85820d4`, one is deferred:
- API: the caller's-word-first effort order regressed GLM-5 (its template never rejects, so "low" became Max) → a load-time **dialect probe** (do "low" and "medium" render differently?) decides between the caller's word (Inkling, Qwen3.8) and the GLM mapping; explicit `enable_thinking: true` beats a `none`. Inkling's **special-token framing** (reasoning glued to the answer, tool calls never surfaced) → `MarkerDialect`/`MarkerFilter` translate the marker ids into `<think>…</think>` and Hermes `<tool_call>` blocks on every response path; the single-stage engine path now emits per-token chunks with ids so the translation has something to read. `tojson` refuses `indent` + `separators`.
- Engine: rewind slack bounded against a write high-water mark (consecutive truncates could read overwritten rows); O(1) reset with the invariant proven; NaN-safe top-k; one `Head::logits` (`x / mup`, no 1-ULP drift at mup 24); `StFile::bf16_bits` (no float32 detour, ~4× less transient memory on the edge tables); overlapped expert reads after routing (`CASCADIA_INKLING_SEQ_READS=1` for serial); manifest fields `route_scale`, `log_scaling_alpha`, `logits_mup_width_multiplier` required and zero intermediates rejected; shared `peek_arch` / tokenizer / max_seq helpers across the three Rust-shell arms.
- Exporter: direct write when every source shard is present (no staging round-trip), `--tiny` refuses to wipe a foreign directory without `--force`, fixtures regenerated with a load-bearing relative-position bias.
- Deferred: row-batched projection GEMV for prefill (inherited from the GLM path; matters once experts are page-cache-hot) — on the follow-up list.

## Expert-parallel dispatch (added on request)
Beside the layer pipeline, the family runs as a **driver + expert workers** star: the driver (`--ep-workers a:p,…`, single stage) holds every layer's attention/router/dense/embed/head and no expert weights; worker N of W (`--ep-worker-index N --ep-worker-count W --listen :port`) holds the experts with `id % W == N` for every MoE layer (shared experts dispatched too). Per MoE layer the driver routes locally, dispatches hidden rows + expert ids to the involved workers together (`ExpertDispatch` / `ExpertResult`, appended frame kinds), applies the weights and sums in gate order — **bit-identical** to the single-process forward (asserted over loopback TCP for eager and mmap experts, token-by-token, batched and >256-row prefill; worker errors surface as task errors). Loopback protocol cost ~150–200 µs per MoE layer. On the miner with driver + 3 workers on one box (real export, SSD-paged): identical answers, cold TTFT 42 s (vs 107–176 s), 16-token answer 121 s vs 174–180 s. The sizing note (`docs/perf/INKLING_SCALING.md`) sets the expectation across real boxes: ~1.5–2× a resident pipeline on a cabled LAN, ~5× when experts page from NVMe, 20× slower over a relay.

## Resident regime: Mac Pro 2019 (1.5 TB, 28c/56t Cascade Lake, macOS 12.7) — 2026-09-11
Same binary and prompts as the miner, answers byte-identical. The first resident run (mmap, 56 threads) gave TTFT 39.6 s and **3.0 s/token** — the shell, not memory, was the clock. Four bit-identical schedule fixes (each byte-compared on an 8-layer dump against the serial build) bring it to **TTFT 12.5 s, 0.66 s/token = 1.5 tok/s** single process (64-token answer 53.9 s vs 227 s):
- `perf(inkling): run a token's selected experts concurrently` — the 8 routed+shared experts ran one after another (`CASCADIA_INKLING_SERIAL_EXPERTS=1` restores).
- `perf(inkling): copy an expert's bin only when it is paged out` — the per-token whole-bin copy costs 2× when resident on Linux (miner warm 28 → 14 ms per MoE layer) but still wins cold from SSD (95 vs 110 ms); each expert samples its own residency (`mincore`) per token.
- `feat(inkling): CASCADIA_INKLING_PIN_EXPERTS` — macOS re-faults cached file pages on every touch, so the direct path was *slower* there than the copy; `mlock` the export at load (490 GB in 812 s) fixes it (45 → 33 ms).
- `perf(sparse-moe): size the rayon pool to the physical cores by default` — 56 hyperthreads vs 28 cores on macOS: **33 → 10.4 ms per MoE layer** (miner 48 → 24 indifferent). `RAYON_NUM_THREADS` overrides.
Driver + 3 workers on the same box (pinned, 28 + 3×9 threads): identical text, 0.74 s/token — the star costs ~1.2 ms per MoE layer on loopback. Per-layer tables: `docs/perf/INKLING_SCALING.md` §2/§2a, `docs/architectures/inkling.md`.

## Hardware note
The miner is a correctness platform for this model: decode is SSD-bound (0.05–0.13 tok/s). A record run needs the 512 GB int4 artifact RAM-resident: measured **1.5 tok/s** on the Mac Pro with the pin; ~2.6× off the bandwidth floor, the rest is the shell (bf16 attention GEMVs, per-row dequant, unbatched prefill).

## Follow-ups (not in this PR)
Row-batched prefill GEMV (prefill is per token today: 0.5 s/token resident); int4 attention / fused q,k,v GEMVs; prefix cache for this family (TTFT); hot/cold expert residency for paged hosts; MTP / n-gram speculative decode (rewind slack is in place); multi-box expert-parallel on a cabled LAN; vision/audio inputs.


7i7o (approved):
tACK 👍

PR #155, opened 2026-09-14
feat(inkling): Intel iGPU backends (per-expert, fused MoE, int8 attention, head) + single-box bench: CPU 1.10 / iGPU 1.24 / OpenVINO-native 0.59 tok/s

t8:
## Inkling on the Intel iGPU (Arc B390) through the sparse-MoE engine

Prerequisite: **#157 (CPU changes into `main`) must merge first**, followed by this PR, then #156. Follow the branch-transplant instructions in #157: replay only GPU changes after `fdcc0433` onto updated `main`, rerun checks, and retarget this PR to `main`. Do not merge this PR into the legacy autolab branch. Its current `perf/inkling-panther-autolab` base reflects the original benchmark ancestry. Adds three opt-in OpenVINO backends to the Inkling Rust shell plus the tooling to export their IRs, and measures them the way the autolab measures its record.

### Whole-model single-stream results on tate-07 (campaign-129 protocol)

Full 66-layer model from the 549 GB int4 export, `inkling_decode_bench`, three prompts × three repetitions × 64 tokens, process pinned (High, affinity 65535), the promoted PTL env profile. Score = min over nine samples of decode tok/s.

| configuration | **min** / median / max tok/s | prefill | parity with the bf16 greedy reference |
|---|---|---|---|
| CPU control (record binary, campaign 129 as documented) | **1.098** / 1.127 / 1.200 | 24.6–26.1 s | exact (hash `ce0fbb9a116d3d09`) |
| ours on iGPU: int8 attention IRs (66 layers) + int8 head, Rust bf16 copies released, 16 GiB expert cache | **1.144** / 1.155 / 1.233 | 20.0–22.1 s | 1 prompt exact, 2 part at token 47 / 28 of 64 |
| ours on iGPU, freed RAM to the expert cache (24 GiB) | **1.236** / 1.263 / 1.337 | 20.2–21.3 s | same |
| OpenVINO-native MoE: their fused kernel **and** their on-disk expert offload on layers 2–7 | **0.587** / 0.607 / 0.628 | 43–46 s | 1 exact, 2 part at 29 / 28 |
| OpenVINO fused MoE with layers 2–4 resident on the device | **1.003** / 1.118 / 1.242 | 19–21 s | 2 exact, 1 parts at 29 |

[...]

PR #156, opened 2026-09-16
feat(inkling): add fused iGPU expert routing and full-model topology validation

t8:
## Summary

Route Inkling's selected MoE experts to explicit, capacity-checked CPU/iGPU workers while keeping the complete 66-layer decoder on the driver. Compact fused GPU workers return individual expert outputs; the driver applies original routing weights in original gate order, so changing ownership does not change floating-point reduction order. Optional lossless FP16 replies halve expert tensor traffic while restoring exact FP32 values.

**Depends on #155** and targets its `feat/inkling-igpu` branch. Review/merge order is **#157 (CPU prerequisites into `main`) → #155 (GPU backends) → this PR (distributed expert routing)**. Follow the downstream branch-transplant instructions in #157 before retargeting; changing base alone would bring back the legacy autolab history.

The implementation includes placement/replica validation, fused shard export, two FP16 overflow fixes, strict GPU execution checks, the native full-model validator, and guarded deployment/audit tools. Generated experiment files are excluded from this PR; all original evidence and historical reports are preserved on [archive/inkling-ep-validation-20260915](https://github.com/labscommunity/cascadia/tree/archive/inkling-ep-validation-20260915/docs/perf). Tests generate their own small inputs.

## Validation

| Full-model comparison | Generated choices | Captured tensors | Result |
|---|---:|---:|---|
| Three-NUC CPU EP vs saved single-machine CPU reference | 48 | 3,216 | Bit exact |
| Three iGPU workers vs twelve workers on the same three NUCs | 24 | 1,608 | Bit exact: 51,028,848 floats, with matching payload SHA256 |

The complete export is 548,985,140,942 bytes. All 64 MoE layers ran fused GPU experts with zero CPU expert calls/errors; attention, dense layers, embedding and head stayed on CPU. Lossless replies used 1,168,207,872 tensor bytes versus 2,336,415,744 FP32-equivalent bytes.

Hardware runs on Panther Lake alpha/beta/charlie and the reference/build work on tate-07 preserved existing services. The archive retains executable/runtime hashes, GPU profiles, reports and cleanup audits. Runtime source is unchanged from those measurements at `9c59e449`.

- 16 native EP TCP tests, 5 fused unit tests, 4 validator negative tests and Windows builds passed during implementation; targeted clippy passed with existing warnings.
- All **24 Python EP tests pass without the archived files**, including synthetic archive success, changed-payload rejection and missing-prompt rejection.
- Re-extracted archived full-model evidence passes the independent topology auditor with the current tools.

## Limits and review guide

Twelve logical workers on **three physical hosts** were tested; twelve physical machines still need qualification. GPU choices match the independent FP16 CPU reference 24/24 and the original BF16 CPU reference 22/24; CPU/GPU numerical parity remains unresolved. These short, instrumented SSD-streaming runs do not establish production quality, concurrency, fault tolerance or 25 tok/s.

Start with [expert routing](https://github.com/labscommunity/cascadia/blob/8c919d73f8fafc9c961bde8ab513307351d090cd/docs/perf/INKLING_EXPERT_ROUTING.md) and [full-model validation/reproduction](https://github.com/labscommunity/cascadia/blob/8c919d73f8fafc9c961bde8ab513307351d090cd/docs/perf/INKLING_FULL_MODEL_EP_VALIDATION.md), then `ep.rs`, `ep_fused.rs`, `ep_placement.rs`, `dist.rs`, and the validator/operator tests. The validation guide shows how to extract and independently audit the separate evidence archive.

## Checklist

- [ ] Full workspace `just check` — not rerun; targeted and hardware validation is listed above. Standard CI targets `main` and does not run automatically for this stacked base.
- [ ] Conventional Commits throughout — some existing experiment commits have descriptive subjects without a prefix; shared history has not been rewritten.
- [x] OV changes tested on real Intel hardware — Panther Lake hosts above; exact runtime DLL hashes/profiles are archived and the upstream OpenVINO environment is described in #155.


7i7o (approved):
tACK 👍

PR #157, opened 2026-09-16
perf(inkling): land CPU prerequisites for iGPU and expert routing

t8:
## Summary

Land the CPU runtime prerequisites used by #155 and #156 directly on current `main`, without importing the autolab history or experiment output. This is the missing first PR in the Inkling GPU/expert-routing stack.

The change contains **15 native source/test files and one guide**: mapped BF16 embeddings, owned packed shared experts, bounded reusable reads/cache, pipelined and causal prediction reads, optional row scheduling, and correctness-aware benchmark tools. It preserves the newer validation, encapsulation and transport fixes from #154. Read/cache/prediction options remain opt-in; predictions prefetch bytes without replacing the actual router's expert choices.

The extracted native delta is `d690bf53..fdcc0433`, applied onto `main` at `d2036387`. CUDA/export orchestration, GPU backends, deployment scripts, logs, snapshots, generated placements and benchmark artifacts are excluded. Historical source/results remain on the existing archive branches and in the linked reproduction record.

## Required merge order

1. **Merge this CPU prerequisites PR into `main` first.** Complete its review and CI before advancing the GPU stack.
2. **Transplant the GPU-only changes for #155 onto updated `main`, then retarget and merge #155.** The old CPU/autolab boundary is `fdcc043370ebc0dc56da414e94edf5b037f8b778`; retain only the GPU commits after that boundary from `feat/inkling-igpu`. Resolve against current `main` APIs, rerun checks, verify that no autolab artifacts entered the diff, and change #155's base to `main`. **Do not merge #155 into `perf/inkling-panther-autolab`.**
3. **After #155 lands, transplant the routing-only changes for #156 onto updated `main`, then retarget and merge #156.** Its original GPU boundary is `f660b51885204837b6bfe28fb44049767fb8a5d9`. Preserve the final artifact cleanup from `8c919d73`, account for any subsequent routing commits, rerun checks and change #156's base to `main`.

For each downstream transplant, preserve its current remote tip before rewriting the shared feature branch, and use an explicit lease when updating it. Merely changing the PR base is insufficient: the original branches still contain the autolab ancestry and would expose the experiment history again. #155 and #156 stay unmerged until their respective prerequisite has landed; this PR does not rewrite either branch.

## Validation

On macOS ARM, with an isolated build directory and debug information disabled to limit disk usage:

- `cargo fmt --all -- --check` passed.
- `cargo clippy --locked --workspace --all-targets` passed with warnings.
- `cargo build --locked -p cascadia` passed.
- `cargo test --locked --workspace --all-targets`: **1,004 passed, zero failed, four intentionally ignored**. The existing e2e harness requires the CLI at `target/debug/cascadia`; that path was linked to the isolated build output for this check.
- Separate default, promoted CPU profile and early-prediction fixture processes: each profile generated **96 tokens**, with identical greedy IDs and full logits hash `3018e9c864366f86`. Both opt-in profiles exercised cache hits and successful prediction reads; the early profile also preserved the guard against enabling third-reader mode at the same time.

These are integration/correctness checks, not new full-model throughput measurements or a new validation of Windows direct I/O / AVX2 on this Mac. The historical full-model Panther Lake CPU record remains [1.1161344306 tok/s with its original source, binary and settings](https://github.com/labscommunity/cascadia/blob/fdcc043370ebc0dc56da414e94edf5b037f8b778/tools/inkling_autolab/INKLING_129_REPRODUCTION.md).

[CPU integration guide and merge instructions](https://github.com/labscommunity/cascadia/blob/dd5155bac8efcf6d7c75545b2098d63e42d30305/docs/perf/INKLING_CPU_PREREQUISITES.md).

## Checklist

- [x] Full local formatting, clippy and workspace-test gate passes (commands above).
- [x] Conventional Commit, one focused extraction commit, authored and committed as t8 without a coauthor.
- [x] OpenVINO hardware requirement assessed: this PR changes CPU code only; GPU implementation and its hardware evidence remain in #155/#156.


7i7o (approved):
tACK 👍

